PR #3, opened 2023-01-30
refactor: Migrate frost-signer & frost-coordinator from stacks-blockchain repo

### LINKED ISSUE #2: Migrate frost-signer & frost-coordinator from stacks-node to this repo
They currently live within the `sBTC` branch of the `stacks-node` repo.

netrome:
Note: I used `tracing` as a drop-in replacement for the custom log macros in the stacks blockchain. I also skipped the `build.rs` environment variable set up because I currently don't see the need for it. 

Closes #2


donpdonp (commented):
Is it possible to git move these files instead of what seems to be making each file a new creation and losing the git history/blame for that file?

sergey-shandar:
> Is it possible to git move these files instead of what seems to be making each file a new creation and losing the git history/blame for that file?

I don't think we can do it across repositories. Inside one repository, AFAIK, Git doesn't really track renames but it tracks the same files. So the strategy can be:
1. copy all files from another repo and commit them as they are.
2. refactor the code in the next commits.

sergey-shandar:
@netrome could we fix the test? For example, add `#[ignore]` if a method is not implemented yet.

donpdonp:
I'd like to create the PR that copies frost signer into trustmachines to preserve most of the attribution, then the Cargo changes etc can be done after that.

donpdonp:
Please remove stacks-signer/ from this PR so it merges cleanly after https://github.com/Trust-Machines/core-eng/pull/5

netrome:
@donpdonp I've rebased on #5, @sergey-shandar I added `#[ignore]` to the unfinished tests in `frost-signer`.

frost-signer/Cargo.toml
```diff
@@ -1,25 +1,13 @@
 [package]
-name = "stacks-signer"
+name = "frost-signer"
 version = "0.0.1"
```
sergey-shandar, line 3:
```suggestion
version = "0.0.1"
license = "GPLv3"
homepage = "https://github.com/Trust-Machines/core-eng"
repository = "https://github.com/Trust-Machines/core-eng"
```

netrome: Thanks! :pray:

sergey-shandar (approved):

frost-signer/Cargo.toml
netrome: Also added to `frost-coordinator`

sergey-shandar:
:shipit:

donpdonp (approved):
looks great. thanks for the tracing log replacement.

PR #5, opened 2023-01-30
copy frost-signer from stacks-blockchain repo

donpdonp:
copy frost-signer from stacks-network/stacks-blockchain/


sergey-shandar (approved):

igorsyl:
@donpdonp is the sbtc branch in stacks-network/stacks-blockchain/ no longer active? Should we close it?

PR #9, opened 2023-01-31
throwaway: Coordinator design draft for early feedback

netrome:
This PR contains a rough design draft for the stacks-coordinator and an implementation plan that will be converted to issues. The purpose of this PR is to make it easy to discuss the requirements/design/plan before I create the issues.

### DIFF
--- a/stacks_coordinator_design.md
+++ b/stacks_coordinator_design.md
@@ -1,0 +1,135 @@
+TODO: Create PR for plan to get review
+
+# sBTC private alpha coordinator design
+## Overview
+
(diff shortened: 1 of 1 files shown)


igorsyl:
Perhaps this shouldn't be a throwaway and we could add to a design directory for reference.

netrome:
> Perhaps this shouldn't be a throwaway and we could add to a design directory for reference.

Sure. To avoid maintaining multiple sources of truth I propose adding the design part of this doc to such a directory, only breaking out the implementation plan as a bunch of issues referring to this doc.

I'll do this tomorrow unless someone proposes an even better suggestion.

PR #18, opened 2023-02-01
Coordinator dkg terminate

donpdonp:
this PR does 3 things

- checks for empty ids_to_wait_for to end the DKG round
- limits exponential backoff timer to 3sec max
- changes the state machine to go back to Idle once DKG round is over (used to require moving into signing phase but that was too tightly coupled)

### DIFF
--- a/frost-coordinator/src/main.rs
+++ b/frost-coordinator/src/main.rs
@@ -1,7 +1,7 @@
 use clap::Parser;
 use hashbrown::HashSet;
 use std::time::Duration;
-use tracing::info;
+use tracing::{debug, info};
 
 use frost_signer::config::Config;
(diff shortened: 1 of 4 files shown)


frost-coordinator/src/main.rs
```diff
@@ -1,6 +1,7 @@
 use clap::Parser;
 use hashbrown::HashSet;
-use tracing::info;
+use std::time::Duration;
+use tracing::{debug, info};
```
netrome, line 4: Nit: Can you move the std import so we separate the standard library imports from third-party crate imports?

netrome (approved):
lgtm, just a minor nit.

PR #32, opened 2023-02-06
make a lib for frost-coordinator, and move Coordinator and related types into it

xoloki:


frost-coordinator/src/coordinator.rs
```diff
@@ -0,0 +1,139 @@
+use std::time::Duration;
```
donpdonp, line 1: nitpick: it'd be nice to have this done with git mv main.rs coordinator.rs, then edit coordinator, to preserve more of the edit history/ownership.

donpdonp: then main.rs can be a new creation.

xoloki: I need to split main.rs into coordinator.rs and main.rs, so git mv doesn't seem right...

xoloki: Seems like I'm gonna lose history in one or the other

donpdonp: agreed, though its 10 lines of minor bits like use and const statements in main vs all of coordinator.rs. its a nitpick so either way is okay with me.

frost-coordinator/src/coordinator.rs
```diff
@@ -0,0 +1,139 @@
...
+    #[command(subcommand)]
+    pub command: Command,
+}
+
+#[derive(clap::Subcommand, Debug)]
+pub enum Command {
+    Dkg,
+    Sign { msg: String },
+    GetAggregatePublicKey,
+}
```
netrome, line 22: Nit: Imo, the Cli type belongs to `main` and not the library. I also think the `.run()` function could be moved back to `main.rs`.

netrome (approved):
LGTM, just a minor optional nit

frost-coordinator/src/coordinator.rs
xoloki: Cli moved

xoloki: Not sure if `run()` belongs in main, it's just dispatching to other Coordinator `fn`

donpdonp (approved):

PR #33, opened 2023-02-06
[frost-signer] Add support for DKG - end steps

donpdonp:
- fill out the DKG_END process
- use 0-based party_id
- add DKG_END unit test


igorsyl:
Thanks @donpdonp! Let's tag issues and PR's without linked issues to the sBTC Alpha.

frost-signer/src/signing_round.rs
```diff
@@ -181,6 +182,50 @@ impl SigningRound {
         }
     }
 
+    pub fn dkg_ended(&self) -> Result<MessageTypes, String> {
+        let parties = self.signer.frost_signer.parties.clone();
+        for mut party in parties {
+            let mut commitments: Vec<PolyCommitment> = vec![];
+            for idx in 0..self.commitments.len() {
+                commitments.push(self.commitments.get(&((idx) as u32)).unwrap().to_owned());
+            }
```
netrome, line 191: Can we store `commintments` in a `BTreeMap` to guarantee the map being sorted by it's key, so that we can just do `self.commitments.values().collect()` here?

frost-signer/src/signing_round.rs
```diff
@@ -181,6 +182,50 @@ impl SigningRound {
...
+            let mut shares: HashMap<usize, Scalar> = HashMap::new();
```
netrome, line 192: Where are we using this variable? I'm just seeing the keys being logged on line 206.

frost-signer/src/signing_round.rs
```diff
@@ -241,17 +295,84 @@ impl SigningRound {
...
+    use hashbrown::HashMap;
+    use p256k1::scalar::Scalar;
+    use rand_core::OsRng;
+
+    use crate::signing_round::{DkgEnd, DkgPublicShare, MessageTypes, SigningRound};
+    use crate::state_machine::States;
+
+    #[test]
+    fn dkg_public_share() {
+        let mut rnd = OsRng::default();
```
netrome, line 327: Nit: (General opinion) In my view, tests should be deterministic - so I'd recommend using a seeded RNG. If we want to fuzz, I'd vouch for randomly generating a seed from an external entropy source such as `OsRng` and logging the seed value for reproducibility in case we hit some unanticipated flakyness.

netrome (approved):
Left some comments that could enhance this, but otherwise approved for further iterations.

frost-signer/src/signing_round.rs
sergey-shandar: BTW: I couldn't find an implementation of pseudo-random generator in `rand_core`.

xoloki: Actually I can't find an impl of that trait in `rand_core`, looks like you'll need to use 

https://rust-random.github.io/rand/rand/rngs/struct.StdRng.html

xoloki: > Nit: (General opinion) In my view, tests should be deterministic - so I'd recommend using a seeded RNG. If we want to fuzz, I'd vouch for randomly generating a seed from an external entropy source such as `OsRng` and logging the seed value for reproducibility in case we hit some unanticipated flakyness.

I've never been convinced by the position that tests should be deterministic.  The arguments seems self contradictory: tests should be deterministic so we can know that a test failure is not due to randomness, but also because test randomness isn't enough for a true fuzz test.   

If you're worried about the former, then you can't really believe the latter.

What's your position @netrome ?

donpdonp: I agree tests should be deterministic. If a test runs on two boxes I expect the conditions of the test to be exactly the same. That makes a test failure reproducible. If the input conditions include randomness, that's essentially a new test case on each run.

xoloki: > I agree tests should be deterministic. If a test runs on two boxes I expect the conditions of the test to be exactly the same. That makes a test failure reproducible. 

That's the benefit to deterministic tests.  The downside is that you will by definition catch fewer bugs.   It's frustrating to wonder why a test fails due to randomness, but if it does so then the code is actually broken.  Deterministic tests hide these corner cases.  

> If the input conditions include randomness, that's essentially a new test case on each run.

#itsafeaturenotabug

netrome: > If the input conditions include randomness, that's essentially a new test case on each run.

This pretty much sums up the problem but also benefits of randomness. Thinking more on this, the important property to me is reproducibility rather than determinism. Although reproducibility requires a level of determinism. I.e. it's okay to have randomness in tests, as long as the randomness is controllable and can be repeated in the event of a test failure.

xoloki: Logging a random seed seems like a reasonable compromise, and gives the benefits of both approaches.

igorsyl: We should have random tests with many iterations to explore new problems. And we should have deterministic tests to avoid regressions.

frost-signer/src/signing_round.rs
donpdonp: further down on line 208 is `party.compute_secret(shares, &commitments) ` that uses shares.

frost-signer/src/signing_round.rs
donpdonp: done. though commitments had to be clone()d to get an owned array of PolyCommitment.

frost-signer/src/signing_round.rs
donpdonp: I tried switching to StdRnd and couldnt get past a trait error `rand_core::RngCore` is not implemented - according to the StdRnd docs it implements `rand::RndCore`. Also it brought up the idea of how the seed gets exposed on error, any failed test report would have to include it. So I put in a `get_rnd()` helper that returns OsRnd as before. Its a single place to put in a new approach later.

frost-signer/src/main.rs
```diff
@@ -52,7 +52,7 @@ fn poll_loop(mut net: HttpNetListen, tx: Sender<Message>, id: u64) {
 
 fn main_loop(config: &Config, net: &HttpNet, rx: Receiver<Message>) {
     let signer_id = config.signer.frost_id;
-    let party_ids = vec![(signer_id * 2 - 1) as usize, (signer_id * 2) as usize]; // take two party/reward slots based on id
+    let party_ids = vec![(signer_id * 2 - 2) as usize, (signer_id * 2 - 1) as usize]; // take two party/reward slots based on id
```
igorsyl, line 55: can we add `assert(signer_id>0 && signer_id <XXX)`

donpdonp: added PARTY_MAX const and checks party_id to be in range.

frost-signer/src/signing_round.rs
netrome: Ah I see. Is the intention to mutate the party in the signing round struct? Because currently `party.compute_secret(...)` will only mutate a clone in this functions scope, and hence the result is discarded unless there's an error.

netrome (approved):

igorsyl (approved):
lgtm

Issue #35, opened 2023-02-07
DKG integration test


donpdonp:
duplicate the communication in running these 5 binaries, in an integration test in rust.
```
relay-server $ cargo run
frost-signer $ cargo run -- --id 1
frost-signer $ cargo run -- --id 2
frost-signer $ cargo run -- --id 3
frost-coordinator $ cargo run -- dkg
```
this depends on each piece functioning as a library.

igorsyl:
@sergey-shandar could you publish the working PR? Or link it to this issue if published already?

PR #36, opened 2023-02-07
relay-server: library API test

sergey-shandar:
Partly #35, #25


frost-test/src/lib.rs
```diff
@@ -0,0 +1,146 @@
+mod sync_test;
+mod v1;
+
+#[cfg(test)]
+mod tests {
```
netrome, line 5: Given that the whole crate is a test, is it necessary to have a nested test module? Couldn't we just do `#![cfg(test)]` on the top of this file and add all tests in the `lib` module?

frost-test/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+            let party = &parties[signers[i]];
+            sigs.push(SignatureShare {
+                id: party.id.clone(),
+                z_i: party.sign(&msg, &signers, nonces),
+                public_key: party.public_key.clone(),
+            });
+        }
+        sigs
+    }
+    */
```
netrome, line 79: Is this code copy-pasted from FROST?

frost-test/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    #[test]
+    fn pure_frost() {
```
netrome, line 82: Code organization nit: I prefer to read the most significant functions (the tests in this case) first, and having any helper functions defined afterwards in their order of occurrence in the code.

frost-test/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+        // let num_nonces = 5;
+        let N: usize = 10;
+        let T = (N * 2) / 3;
+
+        let mut rng = OsRng::default();
```
netrome, line 87: Same comment as in the DKG_END PR: I would like to have tests reproducible, so any randomness should be controlled. We can do this as a follow-up since we're introducing this type of randomness in multiple places now.

frost-test/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+                let B: Vec<Vec<PublicNonce>> = parties
+                    .iter_mut()
+                    .map(|p| p.gen_nonces(num_nonces, &mut rng))
+                    .collect();
+                for p in &mut parties {
+                    p.set_group_nonces(B.clone());
+                }
+                sig_agg.set_group_nonces(B.clone());
+            }
+            */
```
netrome, line 143: Will this be cleaned up?

relay-server/src/bin/relay-server.rs
```diff
@@ -0,0 +1,20 @@
...
+        if let Err(e) = f() {
+            eprintln!("IO error: {e}");
+        }
+    }
+}
+
+fn main() {
+    let addr = "127.0.0.1:9776";
+    let listner = TcpListener::bind(addr).unwrap();
+    println!("Listening {addr}...");
```
netrome, line 18: Nit: I recommend adopting `tracing` for logging facilities.

relay-server/src/http/message.rs
```diff
@@ -0,0 +1,99 @@
...
+        let mut read_line = || -> Result<String, Error> {
+            let mut result = String::default();
+            loop {
+                let b = read_byte()?;
+                if b == 13 {
+                    break;
+                };
+                // BUG: in theory, `b` can be non-UNICODE character and the function
+                // can panic.
+                // TODO: use a vector instead of `String`.
```
netrome, line 36: Should we add a follow-up task to handle this?

netrome: Collaboration opinion: `TODO` comments should either reference an existing issue or be handled before merging. We should not leave any dangling todo comments that we don't plan on implementing.

relay-server/src/http/request.rs
```diff
@@ -0,0 +1,177 @@
...
+        assert_eq!(from_utf8(&rm.content), Ok("Hello!"));
+        assert_eq!(read.position(), REQUEST.len() as u64);
+        let mut v = Vec::default();
+        rm.write(&mut Cursor::new(&mut v)).unwrap();
+        const EXPECTED: &str = "\
+            POST / HTTP/1.1\r\n\
+            hello:someThing\r\n\
+            content-length:6\r\n\
+            \r\n\
+            Hello!";
```
netrome, line 123: Does it make sense to factor out a simple helper to construct these HTTP requests?

relay-server/src/server.rs
```diff
@@ -1,143 +1,123 @@
...
-        type Read = Cursor<&'static str>;
-        type Write = Cursor<Vec<u8>>;
-        fn istream(&mut self) -> &mut Self::Read {
-            &mut self.i
-        }
-        fn ostream(&mut self) -> &mut Self::Write {
-            &mut self.o
-        }
-    }
+    use super::Server;
```
netrome, line 73: Local test modules is the only place I advocate for using wildcard imports
```suggestion
    use super::*;
```

netrome (approved):
I've added some questions and general opinions but overall this looks good to me.

frost-test/src/lib.rs
sergey-shandar: I don't have an opinion yet on this. We may remove the project completely and move everything to `tests` directory to one of the project.

frost-test/src/lib.rs
sergey-shandar: yes, and it doesn't work.

frost-test/src/lib.rs
sergey-shandar: yes. May be not in this PR, I can split this PR into several if needed.

frost-test/src/lib.rs
donpdonp: this integration test will need to limit itself to methods exposed by frost-coordinator and frost-signer, and simulate and verify the communication between them.

frost-test/src/v1.rs
```diff
@@ -0,0 +1,23 @@
+#[cfg(test)]
```
donpdonp, line 1: frost-test doesnt need to test frost methods, frost lib has its own tests. my understanding is frost-test is only high level integration tests.

relay-server/src/http/message.rs
donpdonp: this loop of `let u8 = read_byte(); string.push(u8 as char);` I believe will not panic but will instead break any multi-byte utf-8 char - each byte will get re-encoded as a separate utf-8 character.

donpdonp (approved):
why is all of relay-server in this PR? I thought relay-server had already been merged to main. 

why is there an IOStream implementation here? 

Whats here looks fine I'm just not sure why parts of it are here.

relay-server/src/http/message.rs
sergey-shandar: @donpdonp you are right. I've added a unit test and removed the comment.

relay-server/src/http/request.rs
sergey-shandar: it's only a unit test. For unit tests, I prefer verbose, copy-pasted code :-)

frost-test/src/v1.rs
sergey-shandar: Agree, working on it in the next PR.

sergey-shandar:
> why is all of relay-server in this PR? I thought relay-server had already been merged to main.

yes, the PR contains few more utilities which are needed for testing in one proc.

> why is there an IOStream implementation here?

`IoStream` trait is created to mock TcpListner/TcpStream for unit testing.

> Whats here looks fine I'm just not sure why parts of it are here.

PR #38, opened 2023-02-08
feat: Initial stab at stacks-signer design

netrome:
This design is not super compatible with the current version of frost-signer. Ideas for how to bring these two closer are much appreciated.

### DIFF
--- a/stacks-signer/README.md
+++ b/stacks-signer/README.md
@@ -4,7 +4,7 @@ sign sBTC peg out fulfillments from a shared wallet.
 
 Multiple signer instances communicate with each other and a coordinator using an HTTP relay.
 
-The signer interacts with th sBTC contract to read public keys for other signers
+The signer interacts with the sBTC contract to read public keys for other signers
 and the stacks-coordniator.
 
 ```mermaid
@@ -18,9 +18,9 @@ The signer system should contain an event loop which communicates with an HTTP r
 communicates with a signer entity.
 
 ```rust
-pub trait NetworkedSigner {
+pub trait SignerService {
     type RelayServer;
-    type StacksNode; // Shared implementation with frost-coordinator
+    type StacksNode; // Shared implementation with stacks-coordinator
     type Signer: Signer;
 
     fn relay(&self) -> &Self::RelayServer;
@@ -32,9 +32,9 @@ pub trait NetworkedSigner {
 }
 
 pub trait Signer {
-  fn generate_key_shares(&mut self) -> (PublicShares, PrivateShares);
-  fn public_nonce(&self) -> PublicNonce;
-  fn signature_share(&self, msg: &[u8]) -> SignatureShare;
+  fn generate_key_shares(&mut self, round_id: u64) -> (PublicShares, PrivateShares);
+  fn public_nonce(&self, round_id: u64) -> PublicNonce;
+  fn signature_share(&self, round_id: u64, msg: &[u8]) -> SignatureShare;
 }  
 ```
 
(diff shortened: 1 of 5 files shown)


xoloki (approved):
LGTM

stacks-signer/README.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+## Signer event loop
+A rough outline of the signer event loop
+
+```mermaid
+graph TD
+    A[Read configuration from file and sBTC contract] --> B{Next incoming message}
+    B -->|DKG_BEGIN| C[Generate shares]
+    C --> D[Send shares]
```
donpdonp, line 48:
```suggestion
    C --> D[Send shares]
```

donpdonp (approved):

stacks-signer/README.md
```diff
@@ -0,0 +1,102 @@
+# Stacks signer
+The stacks signer is responsible for participating in the FROST signing scheme to
+sign sBTC peg out fulfillments from a shared wallet.
+
+Multiple signer instances communicate with each other and a coordinator using an HTTP relay.
+
+The signer interacts with th sBTC contract to read public keys for other signers
```
igorsyl, line 7: "th sBTC"

stacks-signer/README.md
```diff
@@ -0,0 +1,102 @@
...
+    S[Signer]-->|Get configuration|D[sBTC contract]
+    S-->|DKG+Signing|R[Relay server]
+```
+
+## Signer design
+The signer system should contain an event loop which communicates with an HTTP relay and
+communicates with a signer entity.
+
+```rust
+pub trait NetworkedSigner {
```
igorsyl, line 21: SignerService?

stacks-signer/README.md
```diff
@@ -0,0 +1,102 @@
...
+    type RelayServer;
+    type StacksNode; // Shared implementation with frost-coordinator
```
igorsyl, line 23: ... with _stacks_-coordinator?

stacks-signer/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+    fn relay(&self) -> &Self::RelayServer;
+    fn stacks_node(&self) -> &Self::StacksNode;
+    fn signer(&self) -> &Self::Signer;
+
+    // Provided method
+    fn run(mut self);
+}
+
+pub trait Signer {
```
igorsyl, line 34: These methods need to be aware of the `round_id`.

igorsyl (changes requested):
This is great! The design of the db storing the private keys is important. Ideally, we would an append-only db to avoid risk of deleting keys.

stacks-signer/README.md
netrome: Wha? Ca't yo fil i th missin lettrs whn readin

stacks-signer/README.md
netrome: Yeah that's better!

stacks-signer/README.md
netrome: Correct!

netrome: Updated.

stacks-signer/README.md
netrome: Updated.

stacks-signer/README.md
netrome: Updated :heavy_check_mark:

stacks-signer/README.md
netrome: Good point, updated.

stacks-signer/README.md
xoloki: It's a curious aspect of the human language ability, but it's possible in many languages to leave the vowels out and still be fully intelligible.  Some languages like Hebrew don't even have glyphs for vowels, they use diacritics which are optional.

netrome: Indeed, while other languages like Japanese and Finnish are completely phonetic. And then we have English... where the grapheme to phoneme correlation is at that sweet spot where you motivated to guess how new words should be pronounced, but you're almost always wrong.

PR #41, opened 2023-02-09
use p256k1 from crates and a frost rev with the same dep

xoloki:


igorsyl (approved):
LGTM. 

Curious why does core-eng depend directly on p256k1?

sergey-shandar (approved):

xoloki:
> LGTM.
> 
> Curious why does core-eng depend directly on p256k1?

There's a couple of places where we use scalars and points explicitly

netrome:
> > LGTM.
> > Curious why does core-eng depend directly on p256k1?
> 
> There's a couple of places where we use scalars and points explicitly

 @xoloki if this is necessary, should we consider re-exporting these types within the FROST crate to make it more approachable for new adopters?

xoloki:
> > > LGTM.
> > > Curious why does core-eng depend directly on p256k1?
> > 
> > 
> > There's a couple of places where we use scalars and points explicitly
> 
> @xoloki if this is necessary, should we consider re-exporting these types within the FROST crate to make it more approachable for new adopters?

I think that's an excellent idea.

PR #44, opened 2023-02-09
Use `workspace.dependencies` if possible.

sergey-shandar:
To avoid famous diamond dependency problem, we can use [workspace.dependencies](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -618,22 +618,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "frost"
-version = "0.1.0"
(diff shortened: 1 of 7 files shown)


xoloki (approved):
LGTM

xoloki (changes requested):
You should be able to remove the p256k1 dependency and change the use statements to 

`wtfrost::{Point, PointError, Scalar};`

igorsyl (approved):
lgtm. we should do this in stacks-blockchain as well. @sergey-shandar feel free to add to eng quality discussion link.

Issue #6, opened 2023-01-30
[frost-signer] Add support for signature generation


donpdonp:
once DKG is in place, the next step is to distribute a message to the signers for them to sign.

PR #48, opened 2023-02-10
frost-signer/coordinator message signing

donpdonp:
Implements #6  in signer and coordinator:

NonceRequest/NonceResponse 

SignatureShareRequest/SignatureShareResponse

goal: signed message

in separate terminals run:
```
relay-server $ cargo run
frost-signer $ cargo run -- --id 3
frost-signer $ cargo run -- --id 2
frost-signer $ cargo run -- --id 1
frost-coordinator $ cargo run dkg-sign -- 1 2 3 4

```


frost-coordinator/src/coordinator.rs
```diff
@@ -62,19 +100,147 @@ where
         result
     }
 
-    pub fn sign_message(&mut self, _msg: &str) -> Result<(), Error> {
+    pub fn sign_message(&mut self, msg: &[u8]) -> Result<(), Error> {
```
xoloki, line 103: Should `sign_message` return the group signature?

frost-coordinator/src/coordinator.rs
```diff
@@ -62,19 +100,147 @@ where
...
+                        self.public_nonces.len(),
+                        self.threshold,
+                    );
+                }
+                msg => {
+                    info!("NonceLoop Got unexpected message {:?})", msg.type_id());
+                }
+            }
+
+            if self.public_nonces.len() == self.threshold {
```
xoloki, line 136: The way v1 signer works, it signs for all parties, not just the ones in the signing set.  So if you short circuit this way you will get more party sigs than you want, and will need to discard the ones not in the signing set

xoloki (changes requested):
Couple of changes and comments.  I'll try this branch when I get back to my computer and test some more

igorsyl:
@donpdonp I resolved the conflicts. Feel free to pop off the merge commit if it's broken.

frost-coordinator/src/coordinator.rs
donpdonp: That line is a fmt change. Its from the original branch, but its good to know more about the requirements of sign_message.

https://github.com/Trust-Machines/core-eng/blob/coordinator-aggregate-key/frost-coordinator/src/coordinator.rs#L126

xoloki (approved):
LGTM.  There's still one small thing, we are waiting for nonces from all parties before trying to sign.  This is because `v1::Signer` will return party sigs from all parties.  We can probably fix this with a little munging on either `frost-signer` or `frost-coordinator`, but I think this PR can go in as is.

igorsyl (approved):
lgtm

PR #50, opened 2023-02-10
frost testing: pure FROST test v1, network events for DKG and signing

sergey-shandar:
Partly #25 and #35


frost-test/src/event/dkg/begin.rs
```diff
@@ -0,0 +1,11 @@
+use super::super::Protocol;
```
netrome, line 1: Nit: Prefer absolute imports over relative imports.

frost-test/src/event/dkg/begin.rs
```diff
@@ -0,0 +1,11 @@
...
+
+/// Should be sent by DKG coordinator.
+/// In theory, a signer can participate in multiple DKGs and be
+/// responsible for multiple parties (DKG and signature shares).
+pub struct Begin<P: Protocol> {
+    pub dkg_id: P::DkgId,
+    pub t: u32,
+    /// Has a size of N
+    pub party_to_signer_map: Vec<P::ParticipantId>,
```
netrome, line 10: I'm a bit confused by this name. What is this Vec mapping? If this is a mapping I'd find it more readable to represent it as a map type, unless it's very important for performance reasons to keep it as a vec.

frost-test/src/event/dkg/share.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl<P: Protocol> Share<P> {
+    fn private_index(&self, party_id: u32) -> usize {
+        assert_ne!(self.party_id, party_id);
+        (if party_id < self.party_id {
+            party_id
+        } else {
+            party_id - 1
+        }) as usize
+    }
```
netrome, line 26: Style nit: Can you move the public functions first in the impl block, and move this helper function afterwards?

frost-test/src/event/dkg/share.rs
```diff
@@ -0,0 +1,49 @@
...
+        party_id: u32,
+        public: PolyCommitment,
+        n: u32,
+        map: impl Iterator<Item = (u32, Scalar)>,
+    ) -> Self {
+        let mut result = Self {
+            dkg_id: dkg_id.clone(),
+            party_id,
+            public,
+            private: (n as usize).new_vec(),
```
netrome, line 41: What's the benefit of using `new_vec()` as opposed to `vec![0; n as usize]`?

frost-test/src/new_vec.rs
```diff
@@ -0,0 +1,11 @@
...
+    fn new_vec<T: Default + Clone>(self) -> Vec<T>;
+}
+
+impl NewVec for usize {
+    fn new_vec<T: Default + Clone>(self) -> Vec<T> {
+        let mut result = Vec::with_capacity(self);
+        result.resize(self, T::default());
+        result
+    }
+}
```
netrome, line 11: I suggest removing this trait in favor of calling `vec![0; n]` whenever someone want to initialize an empty vector.

frost-test/src/signer.rs
```diff
@@ -0,0 +1,168 @@
...
+        let n = signer.n;
+        DkgSigner {
+            signer,
+            countdown: n,
+            public: n.new_vec(),
+            private,
+        }
+    }
+    fn share<P: Protocol>(&mut self, dkg_share: &dkg::Share<P>) -> usize {
+        // TODO: check if `self.public[dkg_share.party_id as usze].is_none()`
```
netrome, line 32: Should this be done before merging?

frost-test/src/signer.rs
```diff
@@ -0,0 +1,168 @@
...
+    rng: &mut P::R,
+) -> Vec<Event<P>> {
+    signer
+        .parties
+        .iter()
+        .map(|party| {
+            // Should `get_poly_commitment` be called before `get_shares`?
+            // If yes, it's an anti-pattern https://en.wikipedia.org/wiki/Sequential_coupling.
+            // Should we `signer.get_poly_commitment` instead of `party.get_poly_commitment` in FROST v2?
+            // Proposal: calculate poly commitments at `Signer::new()`.
```
netrome, line 95: I like the proposal.

frost-test/tests/pure_frost.rs
```diff
@@ -0,0 +1,93 @@
...
+    let mut rng = OsRng::default();
+    let mut signers = [
+        v1::Signer::new(&[0, 1], N, T, &mut rng),
+        v1::Signer::new(&[2], N, T, &mut rng),
+        v1::Signer::new(&[3], N, T, &mut rng),
+    ];
+
+    // DKG (Distributed Key Generation)
+    let A = {
+        let A = signers
```
netrome, line 22: Should we rename A to `polynomial_commitments`?

relay-server/src/http/method.rs
```diff
@@ -0,0 +1,42 @@
+use std::collections::HashMap;
```
netrome, line 1: These `relay-server` changes seems unrelated to the FROST interface from what I can tell, which would be a prime candidate for breaking down this PR to two smaller ones.

netrome (approved):
Good stuff! Added some comments and thoughts.

frost-test/src/event/dkg/share.rs
sergey-shandar: Oh, this my personal pain. I would like to avoid macro if there is a good alternative w/o macro.

netrome: Sure, I see you updated this now. I guess I found the "<usize>.new_vec()" call a bit confusing. If we had an extension trait for vectors with the same method named as `Vec::zeros(n: usize)` I think that would read better than both the macro and the `new_vec()` function.

netrome: To skip the `as` conversion, the method could be `fn zeros(n: impl Into<usize>) -> Vec`.

frost-test/tests/pure_frost.rs
sergey-shandar: I keep the names from the original test from FROST. @xoloki , what do you think?

xoloki: There are 3 names we can use to describe this interaction: signer, party, and key.

For v1, a signer is a group is parties, each which controls one key.  Signer is a code entity which is not part of the protocol.

For v2, a signer is a single party, which controls multiple keys.

Since this code is dealing with v1:: Signer, the variable names are correct.

frost-test/src/signer.rs
sergey-shandar: Also @xoloki

frost-test/src/event/dkg/begin.rs
sergey-shandar: done

frost-test/src/signer.rs
sergey-shandar: The code is not completed and not properly tested yet, so I would keep this note for now.

donpdonp (changes requested):
- `frost-test/src` is a re-implementation of frost-signer and frost-coordinator. @sergey-shandar has said the improvements here will go into frost-signer and frost-coordinator.  Should these changes live in a feature branch to merge with frost-signer and frost-coordinator later?
- frost-test/tests/pure_frost.rs is a re-implementation of the frost unit tests, should that be in a PR against frost? 
- frost-test/tests/server.rs looks like the unit test from relay-server/src/server.rs, Does this need to be in frost-test?
- The remaining changes are to relay-server/. Should it have its own PR?

It looks like this PR can instead be three PRs, one against frost-signer/frost-coordinator, one against frost, and one against relay-server.

sergey-shandar:
> * `frost-test/src` is a re-implementation of frost-signer and frost-coordinator. @sergey-shandar has said the improvements here will go into frost-signer and frost-coordinator.  Should these changes live in a feature branch to merge with frost-signer and frost-coordinator later?
> * frost-test/tests/pure_frost.rs is a re-implementation of the frost unit tests, should that be in a PR against frost?
> * frost-test/tests/server.rs looks like the unit test from relay-server/src/server.rs, Does this need to be in frost-test?
> * The remaining changes are to relay-server/. Should it have its own PR?
> 
> It looks like this PR can instead be three PRs, one against frost-signer/frost-coordinator, one against frost, and one against relay-server.

I agree, we will need to make the changes. Will work on this next week unless we need it now.

frost-test/src/signer.rs
xoloki: There's no reason we can't calculate our poly commitment during new.

But we shouldn't store the dkg private shares, they are ephemeral in the protocol.

frost-test/src/event/dkg/begin.rs
sergey-shandar:
```rust
type PartyId = u32;
trait Dkg {
  fn new<R>(rnd: R) -> Self;
  fn get_poly_commitments(&self) -> impl Map<PartyId, PolyCommitments>;
  fn get_shares(&self) -> impl Map<(PartyId, PartyId), Scalar>>;
}
```

donpdonp:
Sounds good. The code organizations/refactorings here will be a big improvement to frost-signer/frost-coordinator.

frost-test/src/signer.rs
sergey-shandar:
```rust
type PartyId = u32;
trait Dkg {
  fn new<R>(rnd: R) -> Self;
  fn get_poly_commitments(&self) -> impl Map<PartyId, PolyCommitments>;
  fn get_shares(&self) -> impl Map<(PartyId, PartyId), Scalar>>;
  ...
}
``` 

```rust
trait Dkg {
  fn new<R>(rnd: R) -> (Self, impl Map<PartyId, PolyCommitments>, impl Map<(PartyId, PartyId), Scalar>>);
  ...
}
```

```rust
trait Dkg {
  fn new<R>(rnd: R) -> (Self, impl Map<PartyId, (PolyCommitments, impl Map<PartyId, Scalar>)>);
  ...
}
```

stjepangolemac (commented):
Didn't spot anything obviously wrong. As someone mentioned it would be much more manageable if this PR was split into 2 (frost-test and relay-server changes).

jferrant:
Should run cargo clippy to get some easy wins/satisfy the new CI.

relay-server/src/http/request.rs
```diff
@@ -49,7 +35,7 @@ impl Message for Request {
 
     fn first_line(&self) -> Vec<String> {
         [
-            self.method.to_owned(),
+            self.method.to_str().to_string(),
```
jferrant, line 38: I assume this is switched to to_str().to_string() for clarity purposes? Not an objection, just curious.

sergey-shandar: The type of `self.method` has been changed to enum `Method`. The `Method` enum has only the `to_str` function but it has no `to_string` function.

sergey-shandar:
This PR will be split to multiple PRs. The first one os #291

Issue #14, opened 2023-02-01
[stacks-coordinator] Stacks transaction construction


netrome:
Add capabilities to the coordinator to construct the mint and burn sBTC stacks transactions. This is a part of the `FeeWallet` in the coordinator design, but it could be broken out to a separate `StacksWallet`.

```rust
pub trait FeeWallet {
    fn mint_sbtc(&self, op: &stacks_node::PegInOp) -> stacks_node::StacksTransaction;
    fn burn_sbtc(&self, op: &stacks_node::PegOutRequestOp) -> stacks_node::StacksTransaction;
    fn set_wallet_address(&self, address: PegWalletAddress) -> stacks_node::StacksTransaction;

    fn fulfill_peg_out(
        &self,
        op: &stacks_node::PegOutRequestOp,
    ) -> bitcoin_node::BitcoinTransaction;
}
```

Concretely, this issue is to provide an implementation of the `mint_sbtc` and `burn_sbtc` functions above.

## Recommended implementation
The current best way we know to construct stacks transactions is through [stacks.js](https://stacks.js.org/), so this task might require some JavaScript interop.

## Dependencies
- #11

Time estimate: 5 days

## Definition of done
The stacks-coordinator has a method to construct sBTC mint and burn transactions.

Let's use https://docs.hiro.so/stacks.js/feature-guides/sign-transactions

```[tasklist]
### Tasks
- [x] Connect Rust to Javascript
- [x] Make sure Stacks.js works with signing transaction.
- [ ] https://github.com/Trust-Machines/core-eng/issues/84
```

netrome:
Note: We should not over-engineer the stacksjs integration. For the alpha release, delivery speed is key. Also, the solution must be such that the coordinator is possible to embed within the stacks-node Docker image. Sync with Jude for more detailed requirements.

sergey-shandar:
@netrome should the `FreeWallet` functions accept `&mut self` instead of `&self` ?

netrome:
> @netrome should the `FreeWallet` functions accept `&mut self` instead of `&self` ?

Yeah I think that's fine. The initial interface is not set in stone, but serves as a hint of what's needed from this component. Feel free to change the input types and the return types as you need. Just keep an eye on `coordinator.rs` so that it fits together with the other components, and it's good to proactively communicate any interface changes of course :).

sergey-shandar:
Proposed structure:

```rust
pub trait StacksWallet {
[...]

PR #57, opened 2023-02-10
An example how to call JS from Rust using Deno executable

sergey-shandar:
First part of #14

Doc preview: https://github.com/Trust-Machines/core-eng/blob/sergey/js/README.md


netrome (approved):
lgtm. Is the idea to later support arbitrary code evaluation by using something like [eval()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval)?

sergey-shandar:
> lgtm. Is the idea to later support arbitrary code evaluation by using something like [eval()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval)?

I was thinking more about like [RPC](https://en.wikipedia.org/wiki/Remote_procedure_call). However, some commands may execute JS code using [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/Function), `eval` = `evil`.

sergey-shandar:
See also https://github.com/hirosystems/clarinet/ for interop.

sergey-shandar:
@netrome, I've introduced some changes in the PR since you reviewed the code. In particular, the protocol message now is `${JSON}\n". It allows ti use `demo test.mjs` as a console.

netrome:
> @netrome, I've introduced some changes in the PR since you reviewed the code. In particular, the protocol message now is `${JSON}\n`. It allows to use `demo test.mjs` as a console.

Cool, I'll re-review later today

.github/workflows/rust.yml
```diff
@@ -16,6 +16,9 @@ jobs:
 
     steps:
     - uses: actions/checkout@v3
+    - uses: denoland/setup-deno@v1
+      with:
+        deno-version: v1.x
```
netrome, line 21: I'd prefer to specify a very specific version to make sure we don't have any CI time bombs.

netrome (approved):
Just one minor opinion.

igorsyl (approved):
lgtm! small and effective!

Issue #61, opened 2023-02-14
[stacks-coordinator] Stacks node integration


netrome:
The Coordinator needs to interact with stacks nodes to

1. Get peg-in and peg-out operations
2. Query the sBTC contract
3. Broadcast stacks transactions

The integration should roughly correspond to the following trait
```rust
pub trait StacksNode {
    fn get_peg_in_ops(&self, block_height: usize) -> Vec<PegInOp>;
    fn get_peg_out_request_ops(&self, block_height: usize) -> Vec<PegOutRequestOp>;
    fn burn_block_height(&self) -> usize;
    fn next_nonce(&self, addr: StacksAddress);
    fn broadcast_transaction(&self, tx: &StacksTransaction);
}
```

## Definition of done
1. The `stacks-coordinator` has an implementation of the `StacksNode` trait which interacts with a stacks node.
2. The interaction is tested with a local instance of a stacks node.

```[tasklist]
### Dependencies
- [ ] https://github.com/stacks-network/stacks-blockchain/issues/3581
```

stjepangolemac:
I can't run stacks-node locally without crashing. This is blocked by https://github.com/stacks-network/stacks-blockchain/issues/3581.

stjepangolemac:
Maybe we don't even need to run mocknet for this one. It seems I can run mainnet locally without issues.

The problem now is that even if I run the stacks-node from the branch in this PR https://github.com/stacks-network/stacks-blockchain/pull/3577 I don't seem to be getting the right data when I ask for `burn_ops`.

```
curl https://stacks-node-api.testnet.stacks.co/v2/burn_ops/2421185/peg_in

/v2/burn_ops/2421185/peg_in% # This is the response
```

Do you have any thoughts @netrome?

stjepangolemac:
I need `StacksTransaction` type with its serialization logic to proceed. Blocked by https://github.com/Trust-Machines/core-eng/issues/14

donpdonp:
Peg-In response from stacks-node
https://github.com/stacks-network/stacks-blockchain/pull/3511

```
{
  "PegIn": [
    {
      "amount": 1337,
      "block_height": 217,
      "burn_header_hash": "7ce0139a3fe356316409fe704dbdb9ac8f1847c168fe68bd913d0df75c359fe1",
      "peg_wallet_address": {
        "Addr32": [
          false,
          "P2TR",
[...]

PR #62, opened 2023-02-14
feat: PegQueue implementation for stacks-coordinator

netrome:
This PR introduces a PegQueue which observes peg-ins and peg-out-requests from a stacks node (integration to be added in #61), and can be queried for the next unprocessed peg operation.


stacks-coordinator/src/peg_queue/sqlite_peg_queue.rs
```diff
@@ -0,0 +1,349 @@
...
+                entry.block_height as i64, // Stacks will crash before the coordinator if this is invalid
+                serde_json::to_string(&entry.op)?,
+                entry.status.as_str(),
+            ],
+        )?;
+
+        Ok(())
+    }
+
+    fn get_singe_entry_with_status(
```
stjepangolemac, line 119: A minor typo, should be `get_single_entry_with_status`

stacks-coordinator/src/peg_queue/sqlite_peg_queue.rs
```diff
@@ -0,0 +1,349 @@
...
+        let mut entry = self.get_entry(txid, burn_header_hash)?;
+
+        entry.status = Status::Acknowledged;
+        self.insert(&entry)?;
+
+        Ok(())
+    }
+}
+
+impl SqlitePegQueue {
```
stjepangolemac, line 77: It feels a bit unnatural to me that the main struct implementation is defined after the trait implementation that uses it. What do you think?

stacks-coordinator/src/peg_queue/sqlite_peg_queue.rs
```diff
@@ -0,0 +1,349 @@
...
+            .expect_get_peg_out_request_ops()
+            .returning(|height| vec![peg_out_request_op(height)]);
+
+        // No ops before polling
+        assert!(peg_queue.sbtc_op().unwrap().is_none());
+
+        // Should cause the peg_queue to fetch 3 peg in ops
+        peg_queue.poll(&stacks_node_mock).unwrap();
+
+        for height in 1..=3 {
```
stjepangolemac, line 294: I would replace `3` with `number_of_simulated_blocks` here as it's already defined. Usually good to make the test fully configurable or not at all.

stjepangolemac (commented):
A few nitpicks

stacks-coordinator/src/peg_queue/sqlite_peg_queue.rs
netrome: Right, I'll update.

stacks-coordinator/src/peg_queue/sqlite_peg_queue.rs
netrome: Ouch, thanks for spotting!

stacks-coordinator/src/peg_queue/sqlite_peg_queue.rs
netrome: Yeah I figured the trait has the public interface whereas the main struct implementation is just particular implementation details. Hmm but a reader interested in the public interface would look at the trait definition, and anyone reading this file is probably more interested in the specific implementation.

I'll update the order of the impl blocks. Good point!

stacks-coordinator/src/peg_queue/sqlite_peg_queue.rs
netrome: :heavy_check_mark:

stacks-coordinator/src/peg_queue/sqlite_peg_queue.rs
netrome: :heavy_check_mark:

stacks-coordinator/src/peg_queue/sqlite_peg_queue.rs
netrome: :heavy_check_mark:

stjepangolemac (approved):
I would maybe move the SQL queries above the parts that are using them but it's not a big deal. Otherwise it looks good!

netrome:
> I would maybe move the SQL queries above the parts that are using them but it's not a big deal. Otherwise it looks good!

I typically order functions after occurrence to present the most significant functionality to a reader first, and helper functions afterwards. I find that it improves the flow when reading code, but that's of course subjective and I'm open to discussing other conventions.

Side note; in my personal projects you can see me taking this principle to the extreme where I put imports on the bottom of files instead of at the top, because as a reader I want to look up imports like references in a scientific paper or an index in a book. It would be very weird to read a paper which starts with the references, yet that's what we often do as programmers. I think this is a leftover habit from languages which require types to be imported before you can refer to them.

igorsyl:
> Side note; in my personal projects you can see me taking this principle to the extreme where I put imports on the bottom of files instead of at the top, because as a reader I want to look up imports like references in a scientific paper or an index in a book. It would be very weird to read a paper which starts with the references, yet that's what we often do as programmers. I think this is a leftover habit from languages which require types to be imported before you can refer to them.

Whoa. add to team guidelines for consideration! **I wonder if we can config rustfmt to do this our core-eng.**

netrome:
> Whoa. add to team guidelines for consideration! **I wonder if we can config rustfmt to do this our core-eng.**

It actually works out of the box with rustfmt and rust-analyzer. Both tools will happily work with the imports no matter where they are put. It would be amazing if rustfmt could enforce it though, but I don't see any options for that in the [docs](https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#imports_layout).

stacks-coordinator/src/peg_queue/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+            Self::PegOutRequest(op) => Some(op),
+            _ => None,
+        }
+    }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("Http network error: {0}")]
+    SqliteError(#[from] rusqlite::Error),
```
sergey-shandar, line 49: IMHO, interface level errors shouldn't depend on implementation `Sqlite`

sergey-shandar (approved):
LGTM, I would prefer clear separation between persistent storage interface and  implementation (rusqlite)

sergey-shandar:
> > I would maybe move the SQL queries above the parts that are using them but it's not a big deal. Otherwise it looks good!
> 
> I typically order functions after occurrence to present the most significant functionality to a reader first, and helper functions afterwards. I find that it improves the flow when reading code, but that's of course subjective and I'm open to discussing other conventions.
> 
> Side note; in my personal projects you can see me taking this principle to the extreme where I put imports on the bottom of files instead of at the top, because as a reader I want to look up imports like references in a scientific paper or an index in a book. It would be very weird to read a paper which starts with the references, yet that's what we often do as programmers. I think this is a leftover habit from languages which require types to be imported before you can refer to them.

Wow! At the same time, a lot of programming languages can't compile without forward declarations. So, you just learn to read from the bottom :-D, And C lang supports reading backwards:
```c
typedef int const* * const ConstantPointerToMutablePointerToConstantInt;
```

stacks-coordinator/src/peg_queue/mod.rs
netrome: I agree. Actually looking at this, this error is implementation specific and doesn't even belong to the interface. I should move this error type to `sqlite_peg_queue.rs` instead of the interface level module. I'll update.

netrome: I've updated this now so no types defined in peg_queue/mod.rs` relate to implementation details. :heavy_check_mark:

netrome:
> LGTM, I would prefer clear separation between persistent storage interface and implementation (rusqlite)

Thanks for reviewing. I've updated so the interface `peg_queue/mod.rs` do not define any implementation specific types.

stacks-coordinator/src/peg_queue/sqlite_peg_queue.rs
```diff
@@ -0,0 +1,461 @@
+use std::path::Path;
+use std::str::FromStr;
+
+use rusqlite::Connection as SqliteConnection;
```
sergey-shandar, line 4: nit: `RusqliteConnection`.

netrome: Sure :ok_hand: updated.

sergey-shandar:
@netrome nit: one more proposal https://github.com/Trust-Machines/core-eng/pull/74

netrome:
> @netrome nit: one more proposal #74

Thanks, I've fast-forwarded to include the proposal.

sergey-shandar:
:shipit:

stacks-coordinator/src/peg_queue/sqlite_peg_queue.rs
igorsyl: Do we have spell check CI check?  @sergey-shandar

sergey-shandar: VSCode has some extensions. I'm not sure about CI, and Google says many of them.

Issue #65, opened 2023-02-14
[frost-test] DKG and Peg-In (OP_RETURN)


igorsyl:
General idea:

- [x] Run frost dkg using in-process frost library "frost-lib" (no network protocol)
- [x] Peg-in: spend a P2PKH utxo and lock it into P2TR output script using the frost public aggregate key.
- [ ] Confirm Peg-in correctness by submitting to bitcoind regtest node

netrome:
@igorsyl it would be very helpful for me reading this if the issue had a well-defined "definition of done" so that I can understand what the goal is.

I understand from the title that this is to write some sort of design doc for a frost/btc integration test, but it would be helpful to know what the test is for and what we want to assert. For example, is the purpose to verify compatibility with the frost signatures and some bitcoin output format?

igorsyl:
> I understand from the title that this is to write some sort of design doc for a frost/btc integration test, but it would be helpful to know what the test is for and what we want to assert. For example, is the purpose to verify compatibility with the frost signatures and some bitcoin output format?

Yes, I updated the parent issue with the requirement: https://github.com/Trust-Machines/core-eng/issues/60

> For example, is the purpose to verify compatibility with the frost signatures and some bitcoin output format
Purpose is to learn how to integrate frost with bitcoin using taproot scripts. It should be a single file that we could share with the Degens team who are building the Decentralized Mining Pool with the frost-net library.

netrome:
> > I understand from the title that this is to write some sort of design doc for a frost/btc integration test, but it would be helpful to know what the test is for and what we want to assert. For example, is the purpose to verify compatibility with the frost signatures and some bitcoin output format?
> 
> Yes, I updated the parent issue with the requirement: #60
> 
> > For example, is the purpose to verify compatibility with the frost signatures and some bitcoin output format
> > Purpose is to learn how to integrate frost with bitcoin using taproot scripts. It should be a single file that we could share with the Degens team who are building the Decentralized Mining Pool with the frost-net library.
> 
> TBH, a design doc may be overkill for this task.
> 
> @netrome

Oh I missed that there was a parent issue for more context.

igorsyl:
Tracked by:

<img width="266" alt="image" src="https://user-images.githubusercontent.com/1248500/219721646-c22a255e-172d-469c-a82a-6794ab3a9d79.png">

netrome:
> Tracked by:
> 
[...]

PR #67, opened 2023-02-15
Design for integration of frost-net and bitcoin

donpdonp:
`docs/frost-btc.md` contains the design


donpdonp:
the design is committed to the branch instead of a comment on the PR or issue #65.

link for easier reference as the Files Changed view doesnt render markdown.
https://github.com/Trust-Machines/core-eng/blob/test-frost-btc/frost-test/docs/frost-btc.md

I'm not sure which is preferable, putting such a design doc in the PR or as a file in the branch itself. A github PR comment is easier to view, but it also get lost easily once the PR is closed.

igorsyl:
Thank you for writing up this clear design doc @donpdonp 
We definitely want to keep the docs in our repo so I support the way you've structured this PR!

frost-test/docs/frost-btc.md
```diff
@@ -0,0 +1,39 @@
+This intergration test runs through the sBTC process 
+of signer coordination, btc peg-in, and btc peg-out.
+
+## Test setup
+The integration test is a rust function that calls into the
+signer and coordinator crates while assuming there is a network
+but not actually simulating one. Each crate's api 
+should be divisible enough to run in a single threaded test,
+meaning it can avoid any crate-internal main-loops waiting on network
+activity.
```
netrome, line 10: Does this imply we should expose a more fine-grained control over the `stacks-coordinator` functionality in its public interface rather than just the ability to start/stop the main event loop?

frost-test/docs/frost-btc.md
```diff
@@ -0,0 +1,39 @@
...
+* user publishes BTC TX
+
+## Part 3 Peg Out
+* coordinator requests and gathers nonces from the signers
+* coordinator generates BTC TX
+  * output 1: P2PKH of user-address
+* coordinator flood-send signtaure request of BTC TX payload
+* signers respond with signature-share
+* coordinator gathers signature-shares from the signers
+* coordinator publishes BTC TX
```
netrome, line 37: This seems like the peg-out fulfillment flow. We should add a step where the user generates a BTC TX to request pegging out.

netrome (changes requested):
Mostly looks good to me. The only missing piece I see is the step where the peg-out request is made.

frost-test/docs/frost-btc.md
stjepangolemac: If we do need to expose more control over it something like this could work:

1. Extract internal loop logic from `run()` into `run_once()`
2. Repeatedly call `run_once()` from the `run()` loop

With a note that `run_once()` shouldn't take self by value so the coordinator can be used or modified after it's called.

netrome: Sounds like a viable approach, there's no reason to take ownership of self in the current run loop either. It's a pattern I typically apply when a struct corresponds to the state of a thread or async task, to mark that the task handle is the main point of interaction with the event loop.

donpdonp: exactly - i expect this test will be most of the code in main() for signer and coordinator. I made an initial attempt and coordinator has two or three main-loops waiting on network activity. If it is dissected into functions that run to completion immediately, and keep the networking loops in main() then that seems maximally testable.

I took a shortcut and tried to put the signer and coordinator (and now the relay-server) into its own thread and using http://localhost for communication. That has made more progress.

frost-test/docs/frost-btc.md
donpdonp: Done

igorsyl:
Sorry to chime in late here. I propose we complete this issue in two steps:

1. Use frost + rust-bitcoin crates **only** (no core-eng repo) for dkg, and bitcoin transaction signing. This is the minimal proof of concept. I refer to this as frost-**lib** + bitcoin. Knowing how to accomplish this is in the critical path for the Alpha release.

3. Write an end-to-end example using the networking protocol available in core-eng repository. I refer to this as frost-**net** + bitcoin. Arguably, this test is not on the critical path for Alpha so let's create a second child issue and pause and prioritize for next milestone.

frost-test/docs/frost-btc.md
netrome: Cool, sounds good! I'll create an issue to update this in `stacks-coordinator` then.

netrome: Created this issue: https://github.com/Trust-Machines/core-eng/issues/73

netrome (approved):
lgtm, I created https://github.com/Trust-Machines/core-eng/issues/73 to update the stacks-coordinator event loop as proposed here.

igorsyl:
@donpdonp you can use this link to view the doc nicely :)

https://github.com/Trust-Machines/core-eng/pull/67/files#diff-d21335a24962c8b56e9d71517844a5ddd0605530ac0854a423c7eb38b583ee61

stjepangolemac (approved):

frost-test/docs/frost-btc.md
```diff
@@ -0,0 +1,43 @@
...
+* coordinator requests and gathers nonces from the signers
+* coordinator generates Peg-Out Fulfilment BTC TX
+  * output 1: OP_RETURN ">" <amount> <signature> <memo>
+  * output 2: P2PKH user-bitcoin-address 
+* coordinator flood-send signtaure request of BTC TX payload
+* signers respond with signature-share
+* coordinator gathers signature-shares from the signers
+* coordinator publishes BTC TX
+
+
```
igorsyl, line 43: @donpdonp Let's add the learnings and a detailed write-up of Schnorr's signature BIP-340 encoding.

donpdonp: I added a note about BIP340's "X only" pubkey. A more complete write-up needs more investigation into how this affects FROST. I'll merge this much and update it again later if necessary.

frost-test/docs/frost-btc.md
```diff
@@ -0,0 +1,47 @@
...
+### Notes
+
+* Why P2TR(Taproot)? The process depends on FROST which generates a Schnorr signtaure. Only a Taproot output (segwit v1) can be spent by an input with a Schnorr signture.
```
igorsyl, line 46: Link to bitcoin source code line would be great!

donpdonp: Here is the bitcoin code
https://github.com/bitcoin/bitcoin/blob/master/src/script/interpreter.cpp#L391
```cpp
    case SigVersion::WITNESS_V0:
        return EvalChecksigPreTapscript(sig, pubkey, pbegincodehash, pend, flags, checker, sigversion, serror, success);
    case SigVersion::TAPSCRIPT:
        return EvalChecksigTapscript(sig, pubkey, execdata, flags, checker, sigversion, serror, success);
```
 `EvalChecksigTapscript` is one of the only places `CheckSchnorrSignature` gets used. The path for WITNESS_V0 does not.

igorsyl: Make sure you use a permalink. Otherwise, the link may go stale if it references to a tag.

PR #75, opened 2023-02-17
Integration of frost-lib and bitcoin

donpdonp:
`frost-test/tests/pure_frost.rs`  contains the integration test described


igorsyl:
> frost-test/tests/pure_frost.rs contains the integration test described in https://github.com/Trust-Machines/core-eng/issues/65

Hi @donpdonp ! Let's avoid adding explicit loops in the issue/PR graph. You can always check the issue associated with a PR in the side bar. I've updated the description of this PR to remove the circular reference.

image

frost-test/docs/frost-btc.md
```diff
@@ -0,0 +1,39 @@
...
+# Integration Test
+
+## Part 1 DKG
+* coordinator signals begin DKG
+* signers generate PolyCommitments/Public key-shares.
+* signers generate Private key-shares.
+* signers flood-send all private and public key-shares
+* signers gather the public and private key-shares they dont already have
+* coordinator gathers all public key-shares
+* coordinator computes peg-wallet-address
```
igorsyl, line 21: Making sure we don't refer to frost-net as it is in the the scope of a different issue https://github.com/Trust-Machines/core-eng/issues/77

frost-test/tests/pure_frost.rs
```diff
@@ -0,0 +1,182 @@
...
+fn signing_round(
+    message: &[u8],
+    T: usize,
+    N: usize,
+    mut rng: &mut OsRng,
+    signers: &mut [v1::Signer; 3],
+    A: Vec<PolyCommitment>,
+) -> Result<Signature, AggregatorError> {
+    // decide which signers will be used
+    let mut signers = [signers[0].clone(), signers[1].clone()];
```
igorsyl, line 110: @sergey-shandar proposed adding a higher level API to frost for use in single-process application which uses a single trait with dkg and sign methods and hides the complexity of tying all this together. This proposal would make this example much more readable!

cc @donpdonp @xoloki

sergey-shandar:
@donpdonp it looks like, you need to run `cargo update` to update `Cargo.lock` and fix build issues. Probably, the branch ` https://github.com/stacks-network/stacks-blockchain/?branch=3493-sbtc-peg-out-wire-format#1f4594b5` is rebased again.

donpdonp:
In `frost-test/tests/frost-btc.rs` the test named `blog-post` uses a known good signature, and the provided tx bytes and the private key. Everything that is necessary to recreate the signature for this transaction, in order to prove rust-bitcoin can be used to sign transactions, in order to get an integration test for a peg-in transaction followed by a peg-out transaction.

After going over nearly every aspect of the transaction and the signing process, the signatures are not equal, so I'm leaving this update. The code is in the PR and the test can be run with the following command if someone wants to look further as it will be necessary to sign transactions with rust-bitcoin.

`$ cargo test blog_post -- --nocapture`

CALC SIG is the calculated signature from rust-bitcoin, and GOOD SIG is the signature copied from the blog post on segwit signatures.

```
CALC SIG (71) 304402204e843144cc97e6e70aedd0dab850f0e243383aab5787aa4941f407af02900cba0220292c0f16fbba36e5b8d648b77fcb5b21f1bfad81a6814b5a9a600038d30c704c01
GOOD SIG (72) 3045022100f8dac321b0429798df2952d086e763dd5b374d031c7f400d92370ae3c5f57afd0220531207b28b1b137573941c7b3cf5384a3658ef5fc238d26150d8f75b2bcc61e701
```

sergey-shandar:
PR #190 should fix CI for the PR.

frost-test/src/bitcoind.rs
```diff
@@ -19,7 +19,7 @@ pub fn bitcoind_rpc(method: &str, params: impl ureq::serde::Serialize) -> serde_
         Ok(response) => {
             let json = response.into_json::<serde_json::Value>().unwrap();
             let result = json.as_object().unwrap().get("result").unwrap().clone();
-            println!("{} -> {}", rpc, result);
+            //println!("{} -> {}", rpc, result);
```
xoloki, line 22: if we don't need this anymore just remove

frost-test/tests/frost-btc.rs
```diff
@@ -1,13 +1,498 @@
...
+    };
+    bitcoin::blockdata::transaction::Transaction {
+        version: 2,
+        lock_time: PackedLockTime(0),
+        input: vec![peg_in_input],
+        output: vec![peg_in_output_0, peg_in_output_1],
+    }
+}
+
+fn two_phase_peg_in(
```
xoloki, line 362: Are we planning on switching to the two phase still?  If not plz remove

frost-test/tests/pure_frost.rs
```diff
@@ -0,0 +1,93 @@
+use hashbrown::HashMap;
+use rand_core::OsRng;
+use wtfrost::{
+    common::PublicNonce,
+    traits::Signer,
+    v1::{self, SignatureAggregator},
+};
+
+#[test]
+fn pure_frost_test() {
```
xoloki, line 10: Plz change this to use  `test_helpers::{dkg, sign}` as in the other test

xoloki (approved):
Couple of thing to remove or change, o/w LGTM

jferrant (approved):
I had a really hard time following, so perhaps adding a few more comments throughout might help with code maintenance/understanding, but from what I can tell...looks good :D

frost-test/tests/frost-btc.rs
donpdonp: i'll remove them here and save for another branch

Issue #15, opened 2023-02-01
[stacks-coordinator] BTC transaction construction


netrome:
Add capabilities to the coordinator to construct peg-out fulfillments. This is a part of the `FeeWallet` in the coordinator design

```rust
pub trait FeeWallet {
    fn mint_sbtc(&self, op: &stacks_node::PegInOp) -> stacks_node::StacksTransaction;
    fn burn_sbtc(&self, op: &stacks_node::PegOutRequestOp) -> stacks_node::StacksTransaction;
    fn set_wallet_address(&self, address: PegWalletAddress) -> stacks_node::StacksTransaction;

    fn fulfill_peg_out(
        &self,
        op: &stacks_node::PegOutRequestOp,
    ) -> bitcoin_node::BitcoinTransaction;
}
```

Concretely, this issue is to provide an implementation of the `fulfill_peg_out` function. The return type should be a representation of a bitcoin transaction which consumes a UTXO from the peg out request. I recommend using [this Transaction type](https://docs.rs/bitcoin/0.29.2/bitcoin/blockdata/transaction/struct.Transaction.html) or the equivalent in the stacks-blockchain repo.

Note: The UTXO that must be consumed is not present in the current input type, so the function signature must be complemented with the required information to construct the transaction.

To keep this issue small and focused, the returned bitcoin transaction will not have any signatures for it's input. That will be tackled in #52

```[tasklist]
### Dependencies
- [x] #11
- [x] #13
- [ ] #60 
- [ ] https://github.com/Trust-Machines/core-eng/issues/78
```

Time estimate: 3 days

## Definition of done
The stacks-coordinator has a method to construct a bitcoin peg out fulfillment transaction, given a peg out request.

donpdonp:
Picking up this ticket again as peg-out is the important bitcoin transaction to build. peg-in comes from the bridge and the OP_RETURN style peg-in is just one phase so it needs no help from the coordinator.

PR #87, opened 2023-02-24
FeeWallet peg_out

donpdonp:
implementation of FeeWallet from #15


stacks-coordinator/src/fee_wallet.rs
```diff
@@ -0,0 +1,16 @@
+use crate::bitcoin_node;
+use crate::stacks_node;
+
+pub trait FeeWallet {
```
sergey-shandar, line 4: It is the same as https://github.com/Trust-Machines/core-eng/blob/main/stacks-coordinator/src/peg_wallet.rs#L8 ?

PR #88, opened 2023-02-24
Integrating stacks-coordinator and stacks-node

stjepangolemac:
Adds a basic client for stacks-node for stacks-coordinator purposes. There are two unknowns I need to understand better:

1. What is the `StacksAddress` nonce and how to get it? @igorsyl 
2. What is the right type for `StacksTransaction`? @sergey-shandar


igorsyl:
@stjepangolemac Nonces: https://docs.stacks.co/docs/stacks-academy/network#nonces

stjepangolemac:
I need the `StacksTransaction` type with it's serialization logic. This is blocked by https://github.com/Trust-Machines/core-eng/issues/14

igorsyl:
> I need the `StacksTransaction` type with it's serialization logic. This is blocked by #14

Please update the project board to reflect this.

stjepangolemac:
Okay so I've used the `StacksTransaction` introduced by @sergey-shandar's PR but I don't know how to serialize it properly for the stacks-node to accept it. It can be serialized to JSON though.

https://docs.hiro.so/api#tag/Transactions/operation/get_raw_transaction_by_id

There's no request examples, only for responses.

I had a call with @sergey-shandar the other day and we talked about using yarpc to serialize the transaction properly. Here's how to call it manually https://github.com/Trust-Machines/core-eng/tree/main/yarpc/js/stacks

There are two questions for me now:

1. How do I call this code programmatically from the stacks-coordinator?
2. What kind of transaction do we expect in the stacks-coordinator's `broadcastTransaction()` and which function do we need from here https://github.com/hirosystems/stacks.js/tree/master/packages/transactions ?

PR #90, opened 2023-02-27
no Cargo.lock

sergey-shandar:
The PR removes `Cargo.lock` from source control. However, CI (GitHub Actions) will publish `Cargo.lock` as an artifact for build reproducibility. The same `Cargo.lock` will also be used in all tests.


donpdonp (approved):
:+1: :+1:

xoloki:
I've never been a fan of checking in Cargo.lock, but I was under the impression that you guys were seriously in favor of it.  Why remove it now?

sergey-shandar:
> I've never been a fan of checking in Cargo.lock, but I was under the impression that you guys were seriously in favor of it. Why remove it now?

One of the main reasons to use `Cargo.lock` in source control is reproducibility. The PR solves this problem by publishing as an artifact, so you can copy it to your machine and reproduce the build.

xoloki:
> > I've never been a fan of checking in Cargo.lock, but I was under the impression that you guys were seriously in favor of it. Why remove it now?
> 
> One of the main reasons to use `Cargo.lock` in source control is reproducibility. The PR solves this problem by publishing it as an artifact, so you can download it to your machine and reproduce the CI build. So we don't need the `Cargo.lock` in source control any more.
> 
> For example https://github.com/Trust-Machines/core-eng/actions/runs/4285686681

Ah, I see.  So you get the benefit of being able to diagnose CI failures without having to constantly update the file manually.

SGTM!

xoloki (approved):
SGTM

PR #114, opened 2023-03-02
Rust analzyer clippy cleanup

jferrant:
I added some allows that should be carefully considered at a later date. (should_implement_trait should not probably be allowed. better solution would be to implement the FromStr trait, but I wanted this to be the minimal amount of changes to prevent Clippy screaming at me)

### DIFF
--- a/frost-signer/src/net.rs
+++ b/frost-signer/src/net.rs
@@ -56,17 +56,14 @@ impl NetListen for HttpNetListen {
         debug!("poll {}", url);
         match ureq::get(&url).call() {
             Ok(response) => {
-                match response.status() {
-                    200 => {
-                        match bincode::deserialize_from::<_, Message>(response.into_reader()) {
-                            Ok(msg) => {
-                                debug!("received {:?}", msg);
-                                self.in_queue.push(msg);
-                            }
-                            Err(_e) => {}
-                        };
-                    }
-                    _ => {}
+                if response.status() == 200 {
+                    match bincode::deserialize_from::<_, Message>(response.into_reader()) {
+                        Ok(msg) => {
+                            debug!("received {:?}", msg);
+                            self.in_queue.push(msg);
+                        }
+                        Err(_e) => {}
+                    };
                 };
             }
             Err(e) => {
--- a/frost-test/src/lib.rs
+++ b/frost-test/src/lib.rs
@@ -8,11 +8,12 @@ mod tests {
     use hashbrown::HashMap;
     use rand_core::{CryptoRng, OsRng, RngCore};
     use wtfrost::{
-        common::{PolyCommitment, PublicNonce},
+        common::PolyCommitment,
         errors::DkgError,
-        v1::{Party, SignatureAggregator, SignatureShare},
+        v1::{Party, SignatureAggregator},
     };
 
+    #[allow(non_snake_case)]
     fn distribute(
         parties: &mut Vec<Party>,
         A: &Vec<PolyCommitment>,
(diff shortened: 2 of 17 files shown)


sergey-shandar (approved):

frost-signer/src/net.rs
```diff
@@ -56,17 +56,14 @@ impl NetListen for HttpNetListen {
...
-                        match bincode::deserialize_from::<_, Message>(response.into_reader()) {
-                            Ok(msg) => {
-                                debug!("received {:?}", msg);
-                                self.in_queue.push(msg);
-                            }
-                            Err(_e) => {}
-                        };
-                    }
-                    _ => {}
+                if response.status() == 200 {
```
donpdonp, line 59: The match is there to make it easy to add other response codes.

frost-test/src/lib.rs
```diff
@@ -8,11 +8,12 @@ mod tests {
...
     use wtfrost::{
-        common::{PolyCommitment, PublicNonce},
+        common::PolyCommitment,
         errors::DkgError,
-        v1::{Party, SignatureAggregator, SignatureShare},
+        v1::{Party, SignatureAggregator},
     };
 
+    #[allow(non_snake_case)]
     fn distribute(
```
donpdonp, line 17: isnt this already in snake_case?

sergey-shandar: It's about `A` parameter

jferrant: Yes, unfortunately there are a few instances where we have single capital letter vars which make clippy unhappy. (`A` specifically here, but `T` and `N` below.)

frost-signer/src/net.rs
jferrant: This is true, but until we do add those response codes, I am more on the side of doing a single if. However, I can change it back and add an allow instead if that is the general consensus :)

frost-test/src/lib.rs
donpdonp: oh Ive been renaming T and N to `threshold` and `total` in my branch. there is probably a better name for A as well.

frost-signer/src/net.rs
donpdonp: I see. if the goal is to get clippy to be quiet, thats fine to leave it.

frost-test/src/lib.rs
jferrant: I can make those changes here or let you merge your branch/remove the allow in it?

jferrant:
I see that we don't run clippy in our pull request CI pipeline. Any opinions on me adding it?

frost-test/src/lib.rs
donpdonp: i can fix later when I merge.

sergey-shandar:
> I see that we don't run clippy in our pull request CI pipeline. Any opinions on me adding it?

Sure. We can merge this PR and then add clippy to CI: https://github.com/Trust-Machines/core-eng/blob/main/.github/workflows/rust.yml#L49

sergey-shandar:
Here's an example https://github.com/Trust-Machines/frost/blob/main/.github/workflows/ci.yml#L51

yarpc/src/read_ex.rs
```diff
@@ -18,8 +18,7 @@ pub trait ReadEx: Read {
         Ok(buf)
     }
     fn read_exact_vec(&mut self, len: usize) -> Result<Vec<u8>, Error> {
-        let mut buf = Vec::with_capacity(len);
-        buf.resize(len, 0);
+        let mut buf = vec![0; len];
```
sergey-shandar, line 21: I'm trying to avoid macros which can be replaced by functions but I'm too lazy to argue with Clippy.

jferrant: Oh sorry, I didn't know this was a preference. XD

sergey-shandar: No, it's my personal preferences. I can live with Clippy default settings.

PR #115, opened 2023-03-02
Add clippy-check to CI pipeline

jferrant:
I just ended up taking the CI from Frost. Not sure if those two lints are valid to be ignored, but I assumed we should be consistent between all our repos.


sergey-shandar (approved):

xoloki (approved):
LGTM

PR #116, opened 2023-03-04
stacks-signer: add cli for private key generation

jferrant:
Added CLI to generate a private key in stacks-signer per this corresponding [https://github.com/Trust-Machines/core-eng/issues/31#issue](issue). I wasn't sure if it should be just to stdout or to a file. I implemented both. Can easily rip out any code if its unnecessary.

I also use the same logging function that is inside frost-signer and am not sure if this is a desirable dependency and if it should either be re-implemented, or perhaps this function moved to a more commonly accessible location.

I originally used wtfrost to generate a random scalar and blockstack-core's b58 function call to convert it to a sanitized string. (see first commit)

However, I also noticed blockstack-core had a Secp256k1PrivateKey function. I currently just call to_hex on the output, but am wondering if this should really be also a sanitized b58 string? Some feedback would be great :D


xoloki:
If you'd rather stick with Scalar, it implements Display which writes out bsse58.  But it doesn't make a huge difference either way.

donpdonp (approved):
lgtm

xoloki (approved):
LGTM

Issue #113, opened 2023-03-02
Serialize Clarity values


sergey-shandar:
See `stacks.js`
See https://github.com/fungible-systems/micro-stacks

igorsyl:
@sergey-shandar please follow project management guidelines here: https://github.com/Trust-Machines/core-eng/pull/122

kantai:
I commented in the zoom chat last week, but if you're looking to serialize Clarity values and parameters from rust, the `stacks-common` and `libclarity` library outputs from the stacks-blockchain repo can provide you with what you need.

`libclarity::vm::type::Value` structs implement `StacksMessageCodec`, but a bunch of other serialization methods too. Those serializations can be deserialized from within a Clarity contract using `from-consensus-buff?` or generated using `to-consensus-buff?`

The Hiro subnets repo has a couple examples of doing this with signatures over data. This contract checks a signature and validates the underlying content by generating the input buffer using `to-consensus-buff?` 

https://github.com/hirosystems/stacks-subnets/blob/master/core-contracts/contracts/multi-miner.clar#L58-L65

That hash is also generated in rust code in the block proposal signer method:

https://github.com/hirosystems/stacks-subnets/blob/master/src/chainstate/stacks/miner.rs#L2331-L2378

sergey-shandar:
PR #130

PR #130, opened 2023-03-07
Serializing Clarity values from Rust to JS.

sergey-shandar:
Issue #113


stjepangolemac (approved):

stacks-coordinator/src/make_contract_call.rs
```diff
@@ -56,6 +57,41 @@ pub struct SignedContractCallOptions {
     pub senderKey: String,
 }
 
+impl SignedContractCallOptions {
+    pub fn new(
+        contract_address: &str,
+        contract_name: &str,
```
jferrant, line 63: Bit of a nit so feel free to ignore: 

I generally prefer that if a function makes an owned copy internally, it should just take it as the function param type so its clear to the caller what the function actually requires which is in this case an owned String rather than a &str.

jferrant (approved):
Aside from the one nit, LGTM.

stacks-coordinator/src/make_contract_call.rs
sergey-shandar: I agree with the point because in this case, we can move a string instead of creating a new one. The only problem is that I would like to use the function with string literals, like `::new("123098091", ...)`. So I will replace the parameter `imp Into<String>`, in this case, it will satisfy both requirements: 1. no copy if parameter is a string. 2. accepting string literals.

PR #131, opened 2023-03-07
bitcoind daemon/rpc tools for integration tests

donpdonp:
(Broken off from PR#75)
utilities to start, stop and communicate with bitcoind


frost-test/src/bitcoind.rs
```diff
@@ -0,0 +1,69 @@
...
+
+use ctrlc;
+use ctrlc::Signal;
+use nix::sys::signal;
+use nix::unistd::Pid;
+use ureq;
+use ureq::serde_json;
+use ureq::serde_json::Value;
+
+const BITCOIND_URL: &str = "http://abcd:abcd@localhost:18443";
```
stjepangolemac, line 14: If we use a random port for every `bitcoind_rpc()` invocation we can run multiple tests in parallel.

> -rpcport=\<port\>
> Connect to JSON-RPC on <port> (default: 8332, testnet: 18332, signet: 38332, regtest: 18443)

I'm not sure if that's something that we'd consider useful but here it is either way.

stjepangolemac (approved):

frost-test/src/bitcoind.rs
igorsyl: Yes, I've wanted this for a while. We'd also want this for stacks-blockchain, but let's just create an issue and tag it with "stacks-labs".

jferrant (approved):

PR #142, opened 2023-03-10
 Stacks coordinator signer cli

jferrant:
I know this looks like a big pull request but most of the changes are just to enable passing of errors up in a cleaner way. The most important changes are in stacks-coordinator and stacks-signer which now call the appropriate frost-coordinator and frost-signer functions when wanting to perform a dkg signing round

If you want to run the signing round its very similar to the frost-coordinator and frost-signer layout:

`cargo run --bin relay-server`

In the stacks-singer repo in sep terminals:
`../target/debug/stacks-signer run --id 1 --config conf/stacker.toml`
`../target/debug/stacks-signer run --id 2 --config conf/stacker.toml`
`../target/debug/stacks-signer run --id 3 --config conf/stacker.toml`

In the stacks-coordinator repo in sep terminal:
`../target/debug/stacks-coordinator --config conf/coordinator.toml --signer-config conf/stacker.toml dkg`


frost-signer/src/config.rs
```diff
@@ -5,23 +5,11 @@ use toml;
...
     pub total_parties: usize,
     pub minimum_parties: usize,
-}
-
-// on-disk format for frost save data
-#[derive(Clone, Deserialize, Default, Debug)]
-pub struct Signer {
-    pub frost_id: u32,
-    pub frost_state_file: String,
+    pub max_party_id: usize,
```
xoloki, line 12: We don't need the state file anymore?  How are we saving the private polynomial/private key share?

frost-signer/src/net.rs
```diff
@@ -123,6 +122,21 @@ pub enum HttpNetError {
...
     NetworkError(#[from] Box<ureq::Error>),
+
+    #[error("Recv Error: {0}")]
+    RecvError(#[from] mpsc::RecvError),
+
+    #[error("Send Error")]
+    SendError,
+
+    #[error("DKG signing error")]
+    DKGSigningError(String),
```
xoloki, line 133: DKG and signing errors are different beasts, might be better to separate them.  But that can wait for another PR.

stacks-coordinator/src/error.rs
```diff
@@ -0,0 +1,31 @@
+use frost_coordinator::coordinator::Error as FrostCoordinatorError;
+use frost_signer::net::HttpNetError;
+
+use crate::peg_queue::Error as PegQueueError;
+
+/// Helper that uses this module's error type
+pub type Result<T> = std::result::Result<T, Error>;
```
xoloki, line 7: I'm really not a fan of the typedef'd Result pattern, especially when it lives in an error:: module (error::Result looks like an error XD ), but it's not worth holding this PR up.

xoloki (approved):
Mostly just comments, nothing worth holding up the merge.  The most serious was the removal of the state file from the Config; we'll definitely need some way of storing state for alpha.  The rest was just future looking stuff, and personal peeves.

frost-signer/src/config.rs
jferrant: I think I just deleted it as it wasn't used anywhere (yet) but I should have just made it _frost_state_file instead. Will add back in.

frost-signer/src/config.rs
```diff
@@ -5,23 +5,11 @@ use toml;
...
-    pub signer: Signer,
-}
-
-#[derive(Clone, Deserialize, Default, Debug)]
-pub struct Common {
-    pub stacks_node_url: String,
+    pub http_relay_url: String,
     pub total_signers: usize,
     pub total_parties: usize,
     pub minimum_parties: usize,
```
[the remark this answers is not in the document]
donpdonp: I agree with the threshold naming comments above. Given there isnt much context for these config var names, I'd like to see something like parties_threshold, othewise id immediately ask myself "threshold of what?" when reading the config file.

stacks-coordinator/src/error.rs
jferrant: This was a common practice at my work to enable Result<()> to be used across a library and it to always imply that the returned error needed to map to that library's own error type. However, if this is not the preference, I will revert it.

frost-signer/src/config.rs
```diff
@@ -5,23 +5,11 @@ use toml;
...
-    pub common: Common,
-    pub signer: Signer,
-}
-
-#[derive(Clone, Deserialize, Default, Debug)]
-pub struct Common {
-    pub stacks_node_url: String,
+    pub http_relay_url: String,
     pub total_signers: usize,
     pub total_parties: usize,
```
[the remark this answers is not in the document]
donpdonp: A  new PR would be ideal since adding a 'keys' concept will likely be more than changing the names.

frost-signer/src/config.rs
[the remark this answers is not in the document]
xoloki: If we're going to future proof, then `keys_threshold` is valid for both `v1` and `v2`.

frost-signer/src/config.rs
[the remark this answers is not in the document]
xoloki: In `v1` keys = parties, so it would probably be just a rename.  But let's get this PR in first!

stacks-coordinator/src/error.rs
xoloki: You can leave it to another PR if you'd like, this is just a personal nit.

xoloki: You can leave it to another PR if you'd like, this is just a personal nit.

frost-signer/src/config.rs
```diff
@@ -31,30 +19,22 @@ pub struct Cli {
...
-    /// Turn debugging information on
+    /// ID associated with signer
     #[arg(short, long)]
-    id: Option<u32>,
+    pub id: u32,
 }
 
 impl Config {
-    pub fn from_file(path: &str) -> Result<Config, String> {
+    pub fn from_path(path: impl AsRef<std::path::Path>) -> Result<Config, String> {
```
donpdonp, line 36: `pub fn from_path(path: impl AsRef<std::path::Path>)`
does this param mean anything that can be converted to a pointer? aka its more flexible than `&std::path::Path`

donpdonp (approved):

frost-signer/src/config.rs
jferrant: This means anything that can be converted into a Path so it would be String, &String, &str, etc.

frost-signer/src/config.rs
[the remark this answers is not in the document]
jferrant: haha you guys are too quick for me. Renamed again to keys_threshold from threshold :)

frost-signer/src/config.rs
jferrant: I added it back in as just an option in the signer.toml. Might be good to add back in more granular config file with a `[common]` and `[signer]` subsection. Or introduce the state file as a command line option per signer similar to the id.

xoloki: LGTM 4now

PR #146, opened 2023-03-10
stacks-coordinator peg-out fulfillment btc tx builder

donpdonp:
after #142 merged, nearly everything changed in stacks-coordinator and it wasnt worth saving the old branch so a new one (and a new PR) was made.


stacks-coordinator/src/peg_wallet.rs
```diff
@@ -29,3 +34,97 @@ pub trait PegWallet {
...
+        todo!()
+    }
+}
+
+pub struct FileBitcoinWallet {}
+
+impl BitcoinWallet for FileBitcoinWallet {
+    fn fulfill_peg_out(&self, op: &PegOutRequestOp) -> BitcoinTransaction {
+        let bitcoin_txid = bitcoin::Txid::from_slice(op.txid.as_bytes()).unwrap();
+        let utxo = bitcoin::OutPoint {
```
jferrant, line 74: Nit: Might be nice to make a `bitcoin::OutPoint::new(txid, vtxindex)` constructor and/or `impl From<bitcoin::OutPoint> for bitcoing::TxIn {...}`

stacks-coordinator/src/peg_wallet.rs
```diff
@@ -29,3 +34,97 @@ pub trait PegWallet {
...
+        let peg_out_input = bitcoin::TxIn {
+            previous_output: utxo,
+            script_sig: Default::default(),
+            sequence: Default::default(),
+            witness: Default::default(),
+        };
+        //let p2wpk = bitcoin::Script::new_v0_p2wpkh(&user_address.wpubkey_hash().unwrap());
+        let peg_out_output_stx = op.recipient.to_bitcoin_tx_out(op.amount);
+        let peg_out_script = Script::from_hex(&peg_out_output_stx.script_pubkey.to_hex()).unwrap();
+        let peg_out_output = bitcoin::TxOut {
```
jferrant, line 87: Nit: Similarly, would be nice to have constructors for these structs. I generally prefer them to direct instantiation.

stacks-coordinator/src/peg_wallet.rs
```diff
@@ -29,3 +34,97 @@ pub trait PegWallet {
...
+
+    fn set_wallet_address(&mut self, _address: PegWalletAddress) -> Result<StacksTransaction> {
+        todo!()
+    }
+}
+
+pub struct FileBitcoinWallet {}
+
+impl BitcoinWallet for FileBitcoinWallet {
+    fn fulfill_peg_out(&self, op: &PegOutRequestOp) -> BitcoinTransaction {
```
jferrant, line 72: Nit: could start doing some passing up of errors here and change to `fn fulfill_peg_out(&self, op: &PegOutRequestOp) -> Result<BitcoinTransaction>` and not have to call unwrap within the function. I always avoid as best as pssible calling unwrap, only using if absolutely necessary(adding a comment of why I am calling unwrap) or just using it in tests.

jferrant (approved):
I just have some suggestions for code cleanup, but they are not blockers for the functionality itself.

stacks-coordinator/src/peg_wallet.rs
donpdonp: I did give that a try. Line 73 is simply translating between one Txid into a Txid from another crate

`impl From<blockstack_lib::burnchains::Txid> for bitcoin::Txid {...}`

"only traits defined in the current crate can be implemented for types defined outside of the crate"

jferrant: Oh yeah you would need to modify within wherever bitcoin::TxId is first declared and this may not want to introduce the dependency from blockstack_lib.

stacks-coordinator/src/peg_wallet.rs
donpdonp: For corner cases, such as construction a struct with 10 params but 9 of them are default(), I agree. In this case for TxOut I'm not sure of the advantages of a constructor.

jferrant: It mostly is just to make code maintenance a little easier. But  def not as important. In this case though, I would have considered making the constructor something like new(peg_out_output_stx) which internally calls the functions necessary to create peg_out_script since it is only this struct that needs them.

stacks-coordinator/src/peg_wallet.rs
donpdonp: return value changed to Result<BitcoinTransaction> and local Error type extended to include the errors thrown.

stacks-coordinator/src/coordinator.rs
```diff
@@ -107,5 +108,36 @@ impl TryFrom<Config> for StacksCoordinator {
...
 }
 
+impl Coordinator for StacksCoordinator {
+    type PegQueue = SqlitePegQueue;
+    type FeeWallet = WrapPegWallet;
+    type StacksNode = LocalhostStacksNode;
+    type BitcoinNode = LocalhostBitcoinNode;
+
+    fn peg_queue(&self) -> &Self::PegQueue {
+        todo!()
```
xoloki, line 118: ok to keep these `todo!`?

xoloki (approved):
LGTM

PR #163, opened 2023-03-14
Merge BIP-340 work back onto peg-out-2

xoloki:
Repeat DKG and nonce rounds until the aggregate public key and random commitment sum both have even y coords.  

Also display wrapped aggregation errors when logging.

There is a frost fix which this branch is pulling in via GitHub main rather than crates.io.  When we are satisfied there are no more frost/p256k1 changes necessary I will do new major semver releases of both.


xoloki:
Manually squashed and cherry picked to stacks-coordinator/peg-out-2

PR #164, opened 2023-03-14
WIP: Split DKG into separate public/private states

xoloki:
Since we have to repeat DKG until we get an aggregate public key with an even y coord, and since we only need to do the public portion to figure out if we have one, don't try to do both at once.  Instead split the state machine into separate flows.


jferrant:
Replaced by: https://github.com/Trust-Machines/core-eng/pull/175

PR #167, opened 2023-03-14
fix fmt and clippy

sergey-shandar:


frost-coordinator/src/coordinator.rs
```diff
@@ -159,22 +159,18 @@ where
...
                 .public_nonces
-                .iter()
-                .map(|(_i, n)| n.nonce.clone())
+                .values()
+                .map(|n| n.nonce.clone())
                 .collect();
-            let (_, R) = compute::intermediate(msg, &ids, &nonces);
-            if R.has_even_y() {
-                info!("R has even y coord: {}", &R);
+            let (_, r) = compute::intermediate(msg, &ids, &nonces);
```
xoloki, line 168: I would allow non snake case names here, `R` has a specific meaning in frost and `r` is specific to schnorr proofs.

xoloki (changes requested):
Change `r` back to `R` and add the allow non snake case directive.

frost-coordinator/src/coordinator.rs
sergey-shandar: Let me fix it

xoloki (approved):
LGTM thanks!

PR #171, opened 2023-03-14
Randomize port of bitcoind util for integration test

### LINKED ISSUE #138: randomize bitcoind regtest rpc port
use a randomize listen port for frost-test/src/bitcoind.rs to support running tests in parallel

empeje:
This PR aims to close the #138 by adding ways to generate random port based on available port that is exists. This is the first time I write rust, coming from C/C++ and JavaScript.

This PR itself doesn't create a meaningful changes because there's no user who invoke the `bitcoind_mine` or `bitcoind_setup` yet.

---

Closes #138


frost-test/src/bitcoind.rs
```diff
@@ -11,7 +12,8 @@ use ureq;
 use ureq::serde_json;
 use ureq::serde_json::Value;
 
-const BITCOIND_URL: &str = "http://abcd:abcd@localhost:18443";
+const BITCOIND_URL: &str = "http://abcd:abcd@localhost:port";
+const MAX_PORT: u16 = 28443;
 
 pub fn bitcoind_rpc(method: &str, params: impl ureq::serde::Serialize) -> serde_json::Value {
     let rpc = ureq::json!({"jsonrpc": "1.0", "id": "tst", "method": method, "params": params});
```
empeje, line 19: BITCOIND_URL in line above should incorporate the port as well

frost-test/src/bitcoind.rs
```diff
@@ -11,11 +12,13 @@ use ureq;
 use ureq::serde_json;
 use ureq::serde_json::Value;
 
-const BITCOIND_URL: &str = "http://abcd:abcd@localhost:18443";
+const BITCOIND_URL: &str = "http://abcd:abcd@localhost:port";
+const MAX_PORT: u16 = 28443;
+static mut USED_PORT: u16 = 0;
 
-pub fn bitcoind_rpc(method: &str, params: impl ureq::serde::Serialize) -> serde_json::Value {
+pub unsafe fn bitcoind_rpc(method: &str, params: impl ureq::serde::Serialize) -> serde_json::Value {
```
donpdonp, line 19: this needs a new approach without the use of `unsafe`

frost-test/src/bitcoind.rs
```diff
@@ -51,13 +58,28 @@ pub fn bitcoind_setup() -> pid_t {
...
-        BITCOIND_URL
+        get_bitcoin_url(&port)
     );
     thread::sleep(Duration::from_millis(500));
     bitcoind_pid
 }
 
-pub fn bitcoind_mine(public_key_bytes: &[u8; 33]) -> Value {
+unsafe fn get_bitcoin_url(port: &u16) -> String {
+    BITCOIND_URL.replace("port", &port.to_string())
```
donpdonp, line 68: if the port is left in the url as a number, it will parse into a standard library object (http::URI) where the port can be changed through methods on the uri object.

donpdonp (changes requested):

frost-test/src/bitcoind.rs
empeje: @donpdonp I have udpated my approach, but I couldn't find a way to have a global mutable variable even with http:URI, so instead I'm using approach where we will generate `Config` and use it as a dependency to call the rpc. Please see the updated code if that's make sense.

frost-test/src/bitcoind.rs
```diff
@@ -11,11 +12,25 @@ use ureq;
...
+}
+
+pub fn bitcoind_rpc(
+    method: &str,
+    params: impl ureq::serde::Serialize,
+    config: Config,
+) -> serde_json::Value {
     let rpc = ureq::json!({"jsonrpc": "1.0", "id": "tst", "method": method, "params": params});
-    match ureq::post(BITCOIND_URL).send_json(&rpc) {
+    match ureq::post(&*get_bitcoin_url(config)).send_json(&rpc) {
```
stjepangolemac, line 33: It might be more idiomatic to use `get_bitcoin_url(config).as_str()`.

frost-test/src/bitcoind.rs
```diff
@@ -51,18 +67,37 @@ pub fn bitcoind_setup() -> pid_t {
...
-        BITCOIND_URL
+        get_bitcoin_url(config)
     );
     thread::sleep(Duration::from_millis(500));
     bitcoind_pid
 }
 
-pub fn bitcoind_mine(public_key_bytes: &[u8; 33]) -> Value {
+fn get_bitcoin_url(config: Config) -> String {
+    format!("{url}:{port}", url = BITCOIND_URL, port = config.port)
```
stjepangolemac, line 77: I would suggest to format this in a different way:

```
format!("{}:{}", BITCOIND_URL, config.port);
```

The reason being that rust-analyzer and clippy still have trouble properly identifying the variables if they're in the format string.

frost-test/src/bitcoind.rs
```diff
@@ -51,18 +67,37 @@ pub fn bitcoind_setup() -> pid_t {
...
+}
+
+fn find_port() -> u16 {
```
stjepangolemac, line 80: As you said you're starting out with Rust here's how this could be written in a more idiomatic way, even though it might be subjective.

```
fn find_port() -> Option<u16> {
    (18443..MAX_PORT).find(|port| port_is_available(*port))
}
```

frost-test/src/bitcoind.rs
```diff
@@ -35,11 +50,12 @@ pub fn bitcoind_rpc(method: &str, params: impl ureq::serde::Serialize) -> serde_
     }
 }
 
-pub fn bitcoind_setup() -> pid_t {
+pub fn bitcoind_setup(config: Config) -> pid_t {
     let bitcoind_child = Command::new("bitcoind")
```
stjepangolemac, line 54: `Command::new()` returns a `Child` that you can `kill()` so there's no need to cast to PID and use a separate function. I suppose something like this could work:

```
pub fn bitcoind_setup(config: Config) -> Child {
  ...
  ctrlc::set_handler(move || {
    child.kill().unwrap();
  });
  ...
```

stjepangolemac (changes requested):

empeje:
Thanks @stjepangolemac I'll update the PR :)

empeje:
Thank you for your feedback @stjepangolemac I have update everything with your feedback.

frost-test/src/bitcoind.rs
```diff
@@ -33,11 +48,12 @@ pub fn bitcoind_rpc(method: &str, params: impl ureq::serde::Serialize) -> serde_
     }
 }
 
-pub fn bitcoind_setup() -> BitcoinPid {
+pub fn bitcoind_setup(config: Config) -> BitcoinPid {
     let bitcoind_child = Command::new("bitcoind")
```
stjepangolemac, line 52: We now have both the `Child` and a PID. It seems this can be simplified to this:

```
let child = ...;
ctrlc::set_handler(move || {
  println!("kill bitcoind pid {:?}", child.id());
  child.kill();
})
```

EDIT: Actually, just realized you return PID and have it kill it on Drop. In that case maybe remove `Child` and use PID for everything? What do you think?

stjepangolemac (approved):

PR #174, opened 2023-03-15
Sign peg out

xoloki:


donpdonp (approved):

stacks-coordinator/src/coordinator.rs
```diff
@@ -110,11 +149,16 @@ pub struct StacksCoordinator {
...
     local_peg_queue: SqlitePegQueue,
     local_stacks_node: NodeClient,
+    pub local_fee_wallet: WrapPegWallet,
 }
 
 impl StacksCoordinator {
     pub fn run_dkg_round(&mut self) -> Result<PublicKey> {
-        Ok(self.frost_coordinator.run_distributed_key_generation()?)
+        let p = self.frost_coordinator.run_distributed_key_generation()?;
+        match PublicKey::from_slice(&p.x().to_bytes()) {
```
jferrant, line 158: Can also replace this match with 
`PublicKey::from_slice(&p.x().to_bytes()).map_err(|e| Error::BitcoinSecp256k1(e))`

stacks-coordinator/src/coordinator.rs
```diff
@@ -86,16 +90,51 @@ trait CoordinatorHelpers: Coordinator {
...
+        let mut comp = bitcoin::util::sighash::SighashCache::new(&fulfill_tx);
+        let taproot_sighash = comp
+            .taproot_signature_hash(
+                1,
+                &Prevouts::All(&[&fulfill_tx.output[0]]),
+                None,
+                None,
+                SchnorrSighashType::All,
+            )
+            .unwrap();
```
jferrant, line 117: I know this could all be done later as a cleanup in general of our unwraps everywhere, but would be nice to have this error mapped so we don't have to do an unwrap()

donpdonp: can you provide an example implementation of this?

PR #177, opened 2023-03-16
Improve test and add CI for clarinet

empeje:
Self-explanatory.


xoloki (approved):
LGTM thanks!

igorsyl (changes requested):
Can we add install script for Mac using brew?

PR #180, opened 2023-03-17
stacks-doctor - a CLI tool for analyzing and debugging Stacks nodes

stjepangolemac:
Introduces `stacks-doctor`, a tool that is supposed to help with analyzing and debugging running node behavior.

At the moment there's only one functional command `stacks-doctor burns` that returns some stats on BTC burning.

```
$ stacks-doctor -n mainnet burns -b 2000
Burn fee stats for last 2000 blocks: min=386 max=590000 mean=327642 avg=288925
```

I'd like some general feedback on code organization, the `burns` command, and other potential commands. I also plan to include log analysis according to the steps @kantai usually does (https://github.com/Trust-Machines/core-eng/issues/165#issuecomment-1473500490).


igorsyl:
Call it stacks-doctor? Since we can use this tool to monitor non-mining nodes too? And the name is consistent with stacks-node, stacks-signer, stacks-coordinator?

stjepangolemac:
Here's some example output:

```
core-eng $ minedoc -h
Tool for debugging running Stacks nodes

Usage: minedoc <COMMAND>

Commands:
  analyze  Analyze miner
  env      Print related environment variables that are set
  help     Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
core-eng $ minedoc env
MINEDOC_RPC_URL=http://foo.bar
MINEDOC_LOG_FILE=/Users/sgolem/Developer/trustmachines/stacks-blockchain/testnet/stacks-node/node.log
MINEDOC_DB_FILE=/dev/null
core-eng $ minedoc analyze   
Analyze miner

Usage: minedoc analyze <COMMAND>

Commands:
  rpc   Use RPC API
  logs  Use logs
  db    Use database
  all   Use all data sources
  help  Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
core-eng $ minedoc analyze logs
No problems detected
```

stjepangolemac:
> Call it stacks-doctor? Since we can use this tool to monitor non-mining nodes too? And the name is consistent with stacks-node, stacks-signer, stacks-coordinator?

Yep that also works, will change the naming.

stacks-doctor/src/cli.rs
```diff
@@ -0,0 +1,74 @@
...
+    // How many recent blocks to take into account
+    #[arg(short, long, default_value_t = 1000)]
+    pub blocks: u64,
+}
+
+#[derive(Subcommand, Clone, Debug)]
+pub enum Commands {
+    /// Analyze miner
+    #[command(subcommand)]
+    Analyze,
```
netrome, line 45: Nit: This is not another subcommand right?
```suggestion
    Analyze,
```

stacks-doctor/src/main.rs
```diff
@@ -0,0 +1,16 @@
...
+use crate::cli::Args;
+
+mod cli;
+mod commands;
+
+fn main() {
+    let args = Args::parse();
+
+    match &args.cmd {
+        cli::Commands::Analyze => todo!(),
```
netrome, line 12: Is this supposed to call `analyze_logs`? Because although the function is quite minimal atm, it's still there and does something at least.

stacks-doctor/src/cli.rs
```diff
@@ -0,0 +1,74 @@
...
+    #[arg(short, long, env = "DOCTOR_RPC_URL")]
+    pub rpc_url: String,
+
+    /// Path to the node log file
+    #[arg(short, long, env = "DOCTOR_LOG_FILE")]
+    pub log_file: PathBuf,
+
+    /// Path to the node directory with all the databases, usually contains a <mode>/ dir such as xenon/
+    #[arg(short, long, env = "DOCTOR_DB_DIR")]
+    pub db_dir: PathBuf,
```
netrome, line 70: Nit: This could be even cleaner if we only required the necessary information for the different subcommands, so that users are not required to provide information that isn't needed. I see a few options of how to achieve this:

1. Make all of these optional, and return an error if a command requires a parameter that has not been provided.
2. Organize sub-commands according to the resource they operate on. For example, a call could be `stacks-doctor db burn` or `stacks-doctor logs analyze`.
3. Add these fields to the actual sub commands where required.

netrome (approved):
Looks good to me. I've just added a few nit comments.

stacks-doctor/src/commands/burns.rs
```diff
@@ -0,0 +1,52 @@
+use rusqlite::OpenFlags;
+
+use crate::cli::{Args, BurnsArgs, Network};
+
+pub fn burns(args: Args, burns_args: BurnsArgs) {
```
jferrant, line 5: NIT: I know its a simple application right now and that panicing within these functions isn't really a big deal, but I would make all these functions return a Result so you don't have the unwraps everywhere.

jferrant (approved):
Just had a single NIT which is not a blocker to a merge. :)

stacks-doctor/src/cli.rs
stjepangolemac: No, it was a subcommand before but I've simplified it, for now.

stacks-doctor/src/main.rs
stjepangolemac: Yep will make it so it calls that. In the previous version I had separate subcommands for analysing logs, databases, RPC API. That seemed too complicated for now.

stacks-doctor/src/cli.rs
stjepangolemac: I thought I already responded to this comment. I definitely had a response written. 🤔

Clap has some limitations in how you can combine parameters from different commands and I believe option 1 is the only option.

The `analyze` command is making this a bit harder than I'd like it to be as there's a lot of ways to analyze a node's behavior. Maybe we should still have a single command but use the data that was provided, making them all parameters of the subcommand and optional.

PR #183, opened 2023-03-17
Issue 181: Stacks Node Integration Testing

jferrant:
Special cased some errors that the stacks-node can return that I think shouldn't cause the coordinator to blow up. Opinions would be good though.


igorsyl:
Don't forget to mark "Ready for review", otherwise the PR stays in "Draft" mode.

stacks-coordinator/src/coordinator.rs
```diff
@@ -57,6 +57,8 @@ pub enum Error {
     /// "Bitcoin Sighash Error"
     #[error("Bitcoin Sighash Error")]
     BitcoinSighash(#[from] SighashError),
+    #[error("Command Receiver Error: {0}")]
+    CommandReceiverError(#[from] std::sync::mpsc::RecvError),
```
netrome, line 61: The receiver should only error if the sender has been dropped ([link](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.recv)). I think it's fine to panic in that case because it should not be possible and is completely out of the callers control. I am okay with the explicit error too. However, I suggest we're clear about the implication:

```suggestion
    #[error("Command sender disconnected unexpectedly: {0}")]
    UnexpectedSenderDisconnect(#[from] std::sync::mpsc::RecvError),
```

stacks-coordinator/src/stacks_node/client.rs
```diff
@@ -30,65 +30,79 @@ impl NodeClient {
...
+
+    fn get_response(&self, route: &str) -> Result<String, StacksNodeError> {
+        let url = self.build_url(route);
+        debug!("Sending Request to Stacks Node: {}", &url);
+        //TODO: have to use .text() for proper error handling  perhaps the stack node should return a JSON response. e.g.:
+        // {
+        //    "error": "error msg"
+        // }
+        //
+        //Ok(self.client.get(&url).send()?.json::<Value>()?)
```
netrome, line 42: Nice suggestion. I am generally not a fan of merging unresolved TODO comments. In my opinion, they should either:

1. Reference an existing issue - so it's clear that we have a plan to tackle the TODO.
2. Be resolved within the PR itself.

Can you create an issue to resolve this and reference it in the TODO comment?

stacks-coordinator/src/stacks_node/client.rs
```diff
@@ -30,65 +30,79 @@ impl NodeClient {
...
-            .and_then(|res| res.json::<Value>())
-            .map(|json| from_value(json["peg_in"].clone()))??)
+        let response = self.get_response(&format!("/v2/burn_ops/{block_height}/peg_in"))?;
+        let failure_msg = format!("Could not find burn block at height {block_height}");
+        let peg_ins = if failure_msg == response {
+            debug!(failure_msg);
+            vec![]
+        } else {
+            let json: Value = serde_json::from_str(&response)?;
+            from_value(json["peg_in"].clone())?
```
netrome, line 56: We're inconsistent between keeping and discarding namespace info here: `serde_json::from_str` followed by a `from_value` call. My personal preference would be to keep the namespacing information in the `from_value` call, or alias it to something more informative like `json_from_value`, and do the same in the `from_str` call. I find `serde_json::from_value` and `serde_json::from_str` to be the most readable option, but at least both of these calls should be consistent.

stacks-coordinator/src/stacks_node/client.rs
```diff
@@ -30,65 +30,79 @@ impl NodeClient {
...
-        let url = self.build_url(&format!("/v2/burn_ops/peg_in/{block_height}"));
-        Ok(self
-            .client
-            .get(url)
-            .send()
-            .and_then(|res| res.json::<Value>())
-            .map(|json| from_value(json["peg_in"].clone()))??)
+        let response = self.get_response(&format!("/v2/burn_ops/{block_height}/peg_in"))?;
+        let failure_msg = format!("Could not find burn block at height {block_height}");
+        let peg_ins = if failure_msg == response {
```
netrome, line 51: This feels brittle, but I don't have a concrete & simple improvement suggestion at the top of my head.

stacks-coordinator/src/stacks_node/client.rs
```diff
@@ -30,65 +30,79 @@ impl NodeClient {
...
+            self.get_response(&format!("/v2/burn_ops/{block_height}/peg_out_request"))?;
+        let failure_msg = format!("Could not find burn block at height {block_height}");
+        let peg_outs = if failure_msg == response {
+            debug!(failure_msg);
+            vec![]
+        } else {
+            let json: Value = serde_json::from_str(&response)?;
+            from_value(json["peg_out_request"].clone())?
+        };
+        Ok(peg_outs)
```
netrome, line 75: I feel there's a lot of duplicated code here we could consolidate with the corresponding peg_in query function.

stacks-coordinator/src/stacks_node/client.rs
```diff
@@ -30,65 +30,79 @@ impl NodeClient {
...
-            .map(|json| {
-                json["nonce"]
-                    .as_u64()
-                    .map(|val| val + 1)
-                    .ok_or_else(|| StacksNodeError::InvalidJsonEntry)
-            })?
+        let entry = "nonce";
+        //TODO: Should next_nonce fail when fails to load the stacks chain tip?
+        // Or should it just print a warning and return an optional for the caller to handle?
+        //let failure_msg = String("Failed to load Stacks chain tip");
```
netrome, line 92: My preference is to return an explicit error here. A Result gives more context than an Option for the caller.

netrome: Imo this should be handled with a backoff-retry loop from the caller.

netrome (changes requested):
Good stuff. I added some opinions/thoughts that you can do as you please with. The TODOs should be resolved before merging.

stacks-coordinator/src/stacks_node/client.rs
jferrant: Yeah didn't plan to merge them. I put them in as TODOs hoping to get some feedback from code review. I will be sure to remove them and make issues on any that people agree should be issues :)

stacks-coordinator/src/stacks_node/client.rs
jferrant: Yeah I hate it. This is why i suggested having the stacks-node api return a json struct with an error field instead of a straight up string.

stacks-coordinator/src/stacks_node/client.rs
jferrant: Will do! I originally had all the to json functionality nicely in the get_response function until I had to do the ugly error checking :(

stacks-coordinator/src/stacks_node/client.rs
jferrant: https://github.com/Trust-Machines/stacks-blockchain/issues/9

stacks-coordinator/src/stacks_node/client.rs
jferrant: I will leave it as is for now and will create an issue to update our error handling from the stacks-node api calls dependent on stacks-network/stacks-blockchain#3629

stacks-coordinator/src/stacks_node/client.rs
netrome: Nice :rocket:

stacks-coordinator/src/stacks_node/client.rs
netrome: Fair enough :ok_hand:

donpdonp (approved):

PR #196, opened 2023-03-24
build peg-out tx

donpdonp:
copy the working peg-out builder from frost_btc test into stacks-coordinator.

### DIFF
--- a/stacks-coordinator/src/peg_wallet.rs
+++ b/stacks-coordinator/src/peg_wallet.rs
@@ -1,9 +1,10 @@
+use serde::Serialize;
+
 use crate::bitcoin_node;
 use crate::bitcoin_wallet::{BitcoinWallet as BitcoinWalletStruct, Error as BitcoinWalletError};
 use crate::stacks_node;
 use crate::stacks_transaction::StacksTransaction;
 use crate::stacks_wallet::{Error as StacksWalletError, StacksWallet as StacksWalletStruct};
-use serde::Serialize;
 use std::fmt::Debug;
 
 #[derive(thiserror::Error, Debug)]
--- a/stacks-coordinator/src/bitcoin_wallet.rs
+++ b/stacks-coordinator/src/bitcoin_wallet.rs
@@ -1,9 +1,7 @@
 use crate::bitcoin_node::BitcoinTransaction;
 use crate::peg_wallet::{BitcoinWallet as BitcoinWalletTrait, Error as PegWalletError};
 use crate::stacks_node::PegOutRequestOp;
-use bitcoin::hashes::hex::{FromHex, ToHex};
-use bitcoin::hashes::Hash;
-use bitcoin::Script;
+use bitcoin::{hashes::Hash, Script, WPubkeyHash};
 
 #[derive(thiserror::Error, Debug)]
 pub enum Error {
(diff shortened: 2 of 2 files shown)


stacks-coordinator/src/peg_wallet.rs
```diff
@@ -97,15 +98,21 @@ impl BitcoinWallet for FileBitcoinWallet {
...
-        let peg_out_output_stx = op.recipient.to_bitcoin_tx_out(op.amount);
-        let peg_out_script = Script::from_hex(&peg_out_output_stx.script_pubkey.to_hex())?;
+        let user_address_hash =
+            bitcoin::hashes::hash160::Hash::from_slice(&op.recipient.bytes()).unwrap();
+        // let peg_out_output_stx = op.recipient.to_bitcoin_tx_out(op.amount);
+        // let peg_out_script = Script::from_hex(&peg_out_output_stx.script_pubkey.to_hex())?;
+        // let peg_out_output = bitcoin::TxOut {
+        //     value: peg_out_output_stx.value,
+        //     script_pubkey: peg_out_script,
+        // };
```
netrome, line 108: Nit: This should be cleaned up.

stacks-coordinator/src/peg_wallet.rs
```diff
@@ -97,15 +98,21 @@ impl BitcoinWallet for FileBitcoinWallet {
...
-            value: peg_out_output_stx.value,
-            script_pubkey: peg_out_script,
+            value: op.amount,
+            script_pubkey: p2wpk,
         };
         Ok(bitcoin::blockdata::transaction::Transaction {
-            version: 0,
+            version: 2,
             lock_time: bitcoin::PackedLockTime(0),
             input: vec![peg_out_input],
```
netrome, line 117: We should consume a UTXO from the peg wallet to fund the peg-out fulfillment. Is there a ticket to do this?

stacks-coordinator/src/peg_wallet.rs
```diff
@@ -97,15 +98,21 @@ impl BitcoinWallet for FileBitcoinWallet {
...
             output: vec![peg_out_output],
```
netrome, line 118: I guess we can't assume the input UTXOs amount to exactly what we need, so we must return the remaining balance (minus miner fee) to the peg wallet in another output.

netrome (changes requested):
Cleanup request (blocking): There's some stale comments which should be cleaned up before merging.

Clarification request (not blocking): I don't see any associated issue with this PR, so it's a bit hard for me to understand the context and plan revolving this, so I have some clarifying questions about some observations on missing functionality here.

stacks-coordinator/src/peg_wallet.rs
donpdonp: that happens earlier in the function
```
       let peg_out_input = bitcoin::TxIn {
            previous_output: utxo,
```

donpdonp: After reading some more, this needs to be reworked. PegOutRequestOp.txid is the btc transaction of the peg-out request itself.

netrome:
Closing as this due to inactivity. The important features will find their way through @jferrant's work on the coordinator.

PR #200, opened 2023-03-24
Default start block to burn block height

jferrant:
Originally was gonna do a warning, but seemed unnecessary and where I wanted to do it added a bunch of complications. I think the info message about checking for peg in and outs from x to y is a good enough indicator.

Note: commit 1 is just to satisfy broken clippy in frost test. Relevant changes are made in commit 2
One notable change: start_block_height previously implied actually "last_seen_block_height" (as one was always added to it before doing any polling). I changed it so that it is the block that polling actually starts from now. I think this makes more sense from a user's standpoint and ensures we never do any polling like so "Looking at block 2 to 1".


soju-drinker (approved):

stacks-coordinator/src/coordinator.rs
```diff
@@ -193,10 +195,15 @@ impl StacksCoordinator {
 
 impl TryFrom<Config> for StacksCoordinator {
     type Error = Error;
-    fn try_from(config: Config) -> Result<Self> {
+    fn try_from(mut config: Config) -> Result<Self> {
+        let local_stacks_node = NodeClient::new(&config.stacks_node_rpc_url);
+        if config.start_block_height.is_none() {
+            // If a user has not specified a start_block_height, begin from the current burn block height
+            config.start_block_height = Some(local_stacks_node.burn_block_height()?);
+        }
```
netrome, line 203:
```suggestion
        config.start_block_height = config.start_block_height.or_else(|| Some(local_stacks_node.burn_block_height()?))
```

stacks-coordinator/src/peg_queue/sqlite_peg_queue.rs
```diff
@@ -41,7 +43,10 @@ pub struct SqlitePegQueue {
 impl TryFrom<&Config> for SqlitePegQueue {
     type Error = Error;
     fn try_from(cfg: &Config) -> Result<Self, Error> {
-        let start_block_height = cfg.start_block_height.unwrap_or(0);
+        //Note start_block_height should always be set by the time the pegqueue is initailized
+        let start_block_height = cfg
+            .start_block_height
+            .ok_or_else(|| Error::MissingStartBlockHeight)?;
```
netrome, line 49: (Beyond the scope of this review): I am not a fan of the sequential coupling here. This logic should receive `start_block_height` as a value and not an Option. We should not use `TryFrom<&Config>` to construct a SqlitePegQueue. Instead we should define a normal constructor for `SqlitePegQueue`. A `SqlitePegQueue::new(...)` method would push the responsibility of unwrapping the config to the caller, instead of introducing a dead error branch as we have here.

jferrant: Actually, due to your earlier comment, this is no longer a dead error branch.

PR #212, opened 2023-03-27
try to connect to bitcoind for 15 seconds

donpdonp:
makes frost-test/ smarter about launching bitcoind. it tries every half second for 15 seconds to connect to the rpc port on bitcoind. the github CI environment is currently failing on this test due to what I suspect is a performance lag which was fast enough to make the test pass previously. The time to startup is added to the log.


frost-test/src/bitcoind.rs
```diff
@@ -48,12 +47,31 @@ pub fn bitcoind_setup() -> BitcoinPid {
...
+                bitcoind_pid, elapsed
+            );
+            BitcoinPid::new(bitcoind_pid)
+        }
+    }
+}
+
+pub fn connectivity_check() -> Result<f32, String> {
+    let now = std::time::SystemTime::now();
+    for _tries in 1..30 {
```
xoloki, line 66: So you loop up to 30 times, waiting 500ms between iterations.  That means you have a max of 15 seconds for bitcoind to launch and answer RPC calls.

I think this needs to be 60s total, so 120 iterations.  CI is *slow*....

xoloki (changes requested):
Increase the total time to 60s and we're good.

xoloki (approved):
LGTM, thanks!

PR #214, opened 2023-03-27
Config files: Rename http_relay_url to stacks_node_rpc_url

jferrant:
This was irking me so quickly changed so its clear what this should actually be (was also my fault it got changed in the first place...must hide my shame)


frost-coordinator/src/lib.rs
```diff
@@ -14,7 +14,7 @@ pub fn create_coordinator(
 ) -> Result<Coordinator<HttpNetListen>, Error> {
     let config = Config::from_path(path)?;
 
-    let net: HttpNet = HttpNet::new(config.http_relay_url.clone());
+    let net: HttpNet = HttpNet::new(config.stacks_node_rpc_url.clone());
```
netrome, line 17: The purpose of the HttpNet is to connect to the signer http relay and not stacks node so from my perspective the old naming made more sense than the proposed change.

In case there's any confusion, note that "Relay" and "Stacks node RPC" are two different entities in the coordinator design:
image

netrome (changes requested):
Sorry to block this PR, but from my perspective this does not look correct. I suspect that there's a confusion between the responsibilities of the signer relay and the stacks node RPC.

If you think I'm the one being confused, poke me and I'll approve.

PR #219, opened 2023-03-29
Judes sbtc-mini contract and clarinet test

radicleart:
Initial pull of Judes contract into core-eng, sets up Clarinet. 

I will continue adding tests on this branch. The PR can stay open until the tests are written but merging it at any time should not hurt anything.

### DIFF
--- a/sbtc-mini/tests/sbtc-mini.test.ts
+++ b/sbtc-mini/tests/sbtc-mini.test.ts
@@ -1,0 +1,18 @@
+import { Clarinet, Chain, Account } from "../utils/deps.ts";
+import { SbtcMini } from '../models/sbtc-mini.model.ts'
+
+Clarinet.test({
+  name: "sbtc-mini: is_authorized_as_stacker() fails if caller is not DAO or extension",
+  fn(chain: Chain, accounts: Map<string, Account>) {
+    // arrange
+    const deployer = accounts.get("deployer")!;
+    const sbtcMini = new SbtcMini(chain, deployer);
+    const user = accounts.get("wallet_1")!;
+
+    // act
+
+    // assert
+    sbtcMini.is_authorized_as_stacker(user.address, 0).result.expectBool(false);
+  },
+});
+
--- a/sbtc-mini/.gitignore
+++ b/sbtc-mini/.gitignore
@@ -1,4 +1,5 @@
 
-settings/Mainnet.toml
-settings/Testnet.toml
+**/settings/Mainnet.toml
+**/settings/Testnet.toml
+.cache/**
 history.txt
--- a/sbtc-mini/Clarinet.toml
+++ b/sbtc-mini/Clarinet.toml
@@ -1,20 +1,24 @@
 
 [project]
 name = "sbtc-mini"
+description = ""
 authors = []
-telemetry = false
+telemetry = true
+cache_dir = "./.cache"
+
+[contracts.sbtc-mini]
+path = 'contracts/sbtc-mini.clar'
+clarity_version = 2
+epoch = 2.1
+
 [repl.analysis]
 passes = ["check_checker"]
-[repl.analysis.check_checker]
-# If true, inputs are trusted after tx_sender has been checked.
-trusted_sender = false
-# If true, inputs are trusted after contract-caller has been checked.
-trusted_caller = false
-# If true, untrusted data may be passed into a private function without a
+check_checker = { trusted_sender = false, trusted_caller = false, callee_filter = false }
+
+# Check-checker settings:
+# trusted_sender: if true, inputs are trusted after tx_sender has been checked.
+# trusted_caller: if true, inputs are trusted after contract-caller has been checked.
+# callee_filter: if true, untrusted data may be passed into a private function without a
 # warning, if it gets checked inside. This check will also propagate up to the
 # caller.
-callee_filter = false
-
-# [contracts.counter]
-# path = "contracts/counter.clar"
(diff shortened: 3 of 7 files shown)


radicleart:
Added a test that requires some data to be set up in the `pox-2` contract before the target code in sbtc-mini.clar can be reached. Clarinet.toml actually targets a copy of the sbtc-mini contract that has been pulled into the test directory.

See the readme and the scripts dir for more details.

setzeus (commented):
Excellent. Read through & everything looks good to me. 

Hadn't see models in testing before so took me a moment to understand the setup but I follow the structure / agree it helps speed up & cover testing surface.

FriendsFerdinand:
I think we should merge this so we can start branching out from this for Clarity-related work. The Clarity Bitcoin library could be a sub-directory in mini-sbtc. Unless any of you think that it should exist as a separate direct/project in the mini-sbtc directory.

FriendsFerdinand:
For any of the functions that verify segwit Bitcoin transactions that hash to a wtxid, we're going to need 2 transactions:
1. The Coinbase transaction that includes the output with the segwit merkle proof.
2. The segwit transaction that we're interested in with the witness data (if we're using the witness data).

The functions in the current version assume that we only need the 1 non-segwit transaction that has to be verified.

sbtc-mini/tests/sbtc-mini.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  fn(chain: Chain, accounts: Map<string, Account>) {
+    // arrange
+    const deployer = accounts.get("deployer")!;
+    const sbtcMini = new SbtcMini(chain, deployer);
+    const user = accounts.get("wallet_1")!;
+
+    // act
+
+    // assert
+    sbtcMini.is_authorized_as_stacker(user.address, 0).result.expectBool(false);
```
friedger, line 19: `sbtcMini.is_authorized_as_stacker(user.address, 0)` is the "act" part of the test, isn't it?

sbtc-mini/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+In this way we can reach 100% of the Clarity code under test.
+
+After cloning the repo run the script;
+
+```bash
+./scripts/copy-contracts.sh
+```
+
+This copies and filters the test contract(s) to the test directory replacing fully qualified external contract calls to
```
friedger, line 23: Why not using fully qualified contract calls?

I don't understand why this script is needed.

sbtc-mini/tests/sbtc-mini.test.ts
```diff
@@ -0,0 +1,59 @@
...
+  fn(chain: Chain, accounts: Map<string, Account>) {
+    // arrange
+    const deployer = accounts.get("deployer")!;
+    const sbtcMini = new SbtcMini(chain, deployer);
+    const pox2 = new Pox2(chain, deployer);
+    const user = accounts.get("wallet_1")!;
+
+    // act
+    const { receipts } = chain.mineBlock([
+      pox2.allow_contract_caller(user, SBTC_MINI) // pox 2.1 passes none for indefinite/continuous stacking
```
friedger, line 34: indefinite permission and continuous stacking are two different things.

Maybe add undefined as a parameter to make the code easier to understand. 
```
pox2.allow_contract_caller(user, SBTC_MINI, undefined)
```

sbtc-mini/tests/sbtc-mini.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    ]);
+    // console.log('receipts: ', receipts)
+
+    // assert
+    assertEquals(receipts.length, 1);
+    receipts[0].result.expectOk().expectBool(true);
+    sbtcMini.is_authorized_as_stacker(user.address, 0).result.expectBool(true);
```
friedger, line 41: This test does not assert the unlimited permission

friedger:
> The Clarity Bitcoin library could be a sub-directory in mini-sbtc.

There is a repo for clarity bitcoin library here: https://github.com/friedger/clarity-bitcoin 
Is that different to what you need?

sbtc-mini/README.md
```diff
@@ -0,0 +1,26 @@
+# sbtc-mini
+
+![ci](https://github.com/Trust-Machines/core-eng)
```
sergey-shandar, line 3: The url point to a repo instead of CI. Do we need this link?

sbtc-mini/models/pox-2.model.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  // ---------------------------------------------------------------------------------
+  // read only methods
+  // ---------------------------------------------------------------------------------
+
+  get_allowance_contract_callers (stacker: string, calling_contract: string): ReadOnlyFn {
+    return this.callReadOnlyFn("get-allowance-contract-callers", [types.principal(stacker), types.principal(calling_contract)]);
+  }
+
+  private callReadOnlyFn(method: string, args: Array<any> = [], sender: Account = this.deployer): ReadOnlyFn {
```
sergey-shandar, line 36: Try to avoid `any`. Use `unknown` instead.
```suggestion
  private callReadOnlyFn(method: string, args: Array<unknown> = [], sender: Account = this.deployer): ReadOnlyFn {
```

jo-tm:
@radicleart can i close this PR or you want to do a new PR with something from here?

radicleart:
Yes, this is out of date now @jo-tm

PR #220, opened 2023-03-29
[stacks-coordinator] cleanup StacksWallet trait to be clearer

jferrant:
Have discovered that FileStacksWallet is not actually required and have removed it.

Due to my confusion, I have changed some function names to be more explicit. If you have suggestions for better ones, please let me know. (mint, burn, and set_bitcoin_wallet_address should not actually be calling these corresponding sbtc contract functions, but should be generating the appropriate StacksTransaction (which would either be already signed or signed before being broadcasted). This returned transaction should then be broadcast with all the necessary information to make the appropriate sbtc call.

I also noticed that the contract function names were not correct inside the mint and burn functions so have updated them to the correct ones listed in the sbtc alpha contract.


stacks-coordinator/src/stacks_wallet.rs
```diff
@@ -2,7 +2,7 @@ use crate::{
     make_contract_call::{
         Error as ContractError, MakeContractCall, SignedContractCallOptions, ANY,
     },
-    peg_wallet::{Error as PegWalletError, PegWalletAddress, StacksWallet},
+    peg_wallet::{Error as PegWalletError, PegWalletAddress, StacksWallet as StacksWalletTrait},
```
netrome, line 5: I'm not too fond of having to add "Trait" or "Struct" as a suffix to the trait or struct names. My preference here is to keep more specific names for the concrete implementations, such as `StacksWalletJs` which is a `StacksWallet` implementation based on `stacks.js`. However, I'm fine with this given the time constraints of sBTC alpha too.

netrome (approved):
Definitely an improvement overall. I added one opinion on the struct name == trait name pattern.

stjepangolemac (approved):
With my level of understanding this look good!

One thing I'd suggest to think about is spacing out dense code so it's easier to scan and locate parts of it. Here's what I mean, below is an example of "dense" code:

```
fn foo() {
  <COMPLEX LINE>
  <COMPLEX LINE>
  <COMPLEX LINE>
  <

    COMPLEX BLOCK

  >
  <
    COMPLEX BLOCK
  >
}
```

It might be easier to follow it if it was spaced a bit:

```
fn foo() {
  <COMPLEX LINE>
  <COMPLEX LINE>
  <COMPLEX LINE>

  <

    COMPLEX BLOCK

  >

  <
    COMPLEX BLOCK
  >
}
```

This is all completely subjective but here is when I usually add whitespace around some code:

- Function call chain is more than 2 lines long when properly formatted
- There's a bunch of variable initializations but they can clearly be grouped by `mut` and non `mut`
- Function is non trivial and can be logically grouped: setup, main logic, return value(s)

stacks-coordinator/src/stacks_wallet.rs
jferrant: K i will merge this pull request as is, but probably update it as part of the next one :)

PR #223, opened 2023-03-30
Build and sign transactions for burn/mint/set-bitcoin-wallet-address

jferrant:
Blocked by https://github.com/stacks-network/stacks-blockchain/issues/3647

Closes https://github.com/Trust-Machines/core-eng/issues/262
Closes https://github.com/Trust-Machines/core-eng/issues/199
Closes https://github.com/Trust-Machines/core-eng/issues/172


stacks-coordinator/src/stacks_transaction.rs
```diff
@@ -18,3 +26,18 @@ pub struct StacksTransaction {
     pub postConditionMode: PostConditionMode,
     pub postConditions: LengthPrefixedList,
 }
+
+impl StacksTransaction {
+    pub fn to_blockstack_transaction(self) -> Result<BlockstackTransaction, Error> {
```
stjepangolemac, line 31: Maybe you're already aware but here are some breadcrumbs about serializing this tx before sending it to the network

https://github.com/stacks-network/stacks-blockchain/blob/73af120f2bc24ce2e4d0dc820956985dfe44eb88/src/chainstate/stacks/transaction.rs#L578-L593

stjepangolemac (approved):

stacks-coordinator/src/stacks_transaction.rs
jferrant: Yeah I had tried doing that earlier, but as far as I can tell, it expects a hash representation of the tx, not the json blob. So not entirely sure if I have done it correctly, but fingers crossed. It seems to work. Just need to add some valid testing data.

stacks-coordinator/src/make_contract_call.rs
```diff
@@ -87,22 +90,23 @@ impl SignedContractCallOptions {
             nonce: None,
```
netrome, line 90: Not sure if this is the problem you're experiencing, but we must query the next nonce from the stacks-node and set it here.

You can get the nonce using `/v2/accounts/[Principal]`

Example response `docs/rpc/api/core-node/get-account-data.example.json`

```
{
  "balance": "0x0000000000000000000000000007a120",
  "locked": "0x0000000000000000000000000007a120",
  "unlock_height": 126,
  "nonce": 2867,
  "balance_proof": "0xabce",
  "nonce_proof": "0xabcd"
}
```

frost-signer/src/util.rs
```diff
@@ -0,0 +1,13 @@
...
+pub fn parse_public_key(public_key: &str) -> ecdsa::PublicKey {
+    ecdsa::PublicKey::try_from(public_key)
+        .expect("failed to parse coordinator_public_key from config")
+}
+
+pub fn parse_public_keys(public_keys: &[String]) -> Vec<ecdsa::PublicKey> {
+    public_keys
+        .iter()
+        .map(|s| ecdsa::PublicKey::try_from(s.as_str()).expect("failed to parse ecdsa::PublicKey"))
+        .collect::<Vec<ecdsa::PublicKey>>()
```
sergey-shandar, line 12:
```suggestion
        .collect()
```

xoloki: That's my fault, I moved this code from distributed copypasta into util funds, and forgot to remove the now unnecessary generic arg to `collect`

sergey-shandar:
@jferrant don't hesitate to fix hard coded values for `make_contract_call` `left == right` tests if needed.

frost-signer/src/util.rs
sergey-shandar: Yeah, I do this all the time. BTW, I use `.collect::<Vec<_>>()` to help the Rust type interference algorithm without specifying too much.

stacks-coordinator/src/make_contract_call.rs
```diff
@@ -66,13 +68,16 @@ pub struct SignedContractCallOptions {
...
 impl SignedContractCallOptions {
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         contract_address: impl Into<String>,
         contract_name: impl Into<String>,
         function_name: impl Into<String>,
         function_args: &[Value],
         anchor_mode: AnchorMode,
         sender_key: impl Into<String>,
+        nonce: u64,
```
sergey-shandar, line 79: Could we add a new function `with_nonce` instead? I would prefer minimize a number of parameters for functions, so don't add optional parameters to the functions signature.

jferrant:
I pulled make_contract_call out entirely, but I could add it back and leave the original generic StacksTransaction as a potential route to explore? I am more in favour of the purely rust approach. I find it much easier to grok.

sergey-shandar (approved):
Awesome! No dependency on unstable and hard to debug JS code.

jferrant:
> Awesome! No dependency on unstable and hard to debug JS code.
I created https://github.com/Trust-Machines/core-eng/issues/244 to try to decouple some of this code as you suggested :)

stacks-coordinator/src/stacks_wallet.rs
```diff
@@ -1,84 +1,357 @@
...
+        op: &PegOutRequestOp,
+        nonce: u64,
     ) -> Result<StacksTransaction, PegWalletError> {
-        Ok(self.call("burn!".to_string())?)
+        let function_name = "burn!";
+
+        // Build the function arguments
+        let amount = Value::UInt(op.amount.into());
+        // Retrieve the address from the Message Signature
+        let pub_key = StacksPublicKey::recover_to_pubkey(op.txid.as_bytes(), &op.signature)
```
netrome, line 214: This is incorrect. The signature in the `PegOutRequestOp` signs the sha512 hash of the amount and the locking script of the peg out request transaction. I realize now that this information is not available to the coordinator... :upside_down_face: so we need to figure out how to get that information here.

Option 1: The coordinator fetches this transaction from a bitcoin node in order to retrieve the missing locking script. Ugh...
Option 2: We extend `PegOutRequestOp` in stacks-node to hold this information directly.

Thinking out loud: :thinking: 
From a security perspective both options rely on trusting information received from a running blockchain node (either Stacks or Bitcoin). Given that we'll interface with trusted nodes we host ourselves, this is fine.

I propose proceeding with Option 2. I can write up an issue for it if you think it makes sense.

netrome (changes requested):
Houston we have a problem (regarding the `PegOutRequestOp`).

stacks-coordinator/src/stacks_wallet.rs
jferrant: I am fine with Option 2. I feel like all information required to do something with a burn_op should be in the burn_op itself.

stacks-coordinator/src/config.rs
```diff
@@ -22,6 +22,8 @@ pub struct Config {
     pub signer_config_path: String,
     pub start_block_height: Option<u64>,
     pub rusqlite_path: Option<String>,
+    /// The network version we are using ('mainnet' or 'testnet'). Default: 'mainnet'
+    pub network: Option<String>,
```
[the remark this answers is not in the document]
jferrant: Good catch! My bad for not doing a cleanup runthrough. Was in such haste to get the darn thing done :P

stacks-coordinator/src/stacks_wallet.rs
```diff
@@ -1,84 +1,357 @@
...
-    contract_address: String,
+    contract_address: StacksAddress,
     contract_name: String,
-    sender_key: String,
+    sender_key: StacksPrivateKey,
+    version: TransactionVersion,
+    address: StacksAddress,
 }
 
 impl StacksWallet {
```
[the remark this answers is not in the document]
jferrant: I agree. Done :D

stacks-coordinator/src/stacks_wallet.rs
netrome: Issue formulated: https://github.com/stacks-network/stacks-blockchain/issues/3647 Let's leave it to the Stacks-side to derive the address of the sender (pub_key) https://github.com/stacks-network/stacks-blockchain/issues/3647

stacks-coordinator/src/coordinator.rs
```diff
@@ -127,15 +131,26 @@ pub trait Coordinator: Sized {
 // Private helper functions
 trait CoordinatorHelpers: Coordinator {
     fn peg_in(&mut self, op: stacks_node::PegInOp) -> Result<()> {
-        let _tx = self.fee_wallet().stacks_mut().build_mint_transaction(&op)?;
-        //self.stacks_node().broadcast_transaction(&tx);
+        let nonce = self
+            .stacks_node()
+            .next_nonce(self.fee_wallet().stacks().get_address())?;
```
jferrant, line 136: @netrome is this actually correct? To always grab the nonce using the stacks wallet address?

netrome: I think so. We ask the stacks node for the next nonce for this particular address, which it keeps track of.

soju-drinker: Can we write this in a comment for future usage

stacks-coordinator/src/stacks_wallet.rs
jferrant: Pulled in the change and updated the tests :) had to update the cargo file to point to "next"

stacks-coordinator/src/coordinator.rs
```diff
@@ -217,19 +236,53 @@ impl TryFrom<Config> for StacksCoordinator {
...
+        // 1. No aggregate public key stored in persitent storage anywhere
+        // 2. no address already set in sbtc contract (get-bitcoin-wallet-address)
+        frost_coordinator.run_distributed_key_generation()?;
+        let public_key = bitcoin::PublicKey::from_slice(
+            frost_coordinator
+                .get_aggregate_public_key()?
+                .compress()
+                .as_bytes(),
+        )?;
+        let address = bitcoin::Address::p2wpkh(&public_key, bitcoin_network)?;
```
netrome, line 259: This should be a p2tr address, since we're using Schnorr signatures.

netrome (changes requested):
Good stuff. There's a p2wpkh that I believe should be a p2tr. Please update it and I'll approve.

stacks-coordinator/src/coordinator.rs
jferrant: There are quite a few places where I make this mistake. Will update.

stacks-coordinator/src/peg_wallet.rs
```diff
@@ -35,19 +43,20 @@ pub trait BitcoinWallet {
         &self,
         op: &stacks_node::PegOutRequestOp,
     ) -> Result<bitcoin_node::BitcoinTransaction, Error>;
+    /// Gets the BTC address for the wallet
+    fn get_address(&self) -> &BitcoinAddress;
```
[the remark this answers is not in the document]
netrome: @jferrant any opinions?

stacks-coordinator/src/coordinator.rs
netrome: I see this is resolved now :heavy_check_mark:

stacks-coordinator/src/peg_wallet.rs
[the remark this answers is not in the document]
jferrant: I have no opinion for or against :) will change.

jferrant:
> I think you missed my nit comment about `get_address`. It's not a blocker for merging though.

I saw, but I did forget :P Addressed now (no pun intended...)

PR #238, opened 2023-03-31
Sign and verify `FROST` messages

### LINKED ISSUE #229: [P-02] Sign and validate FROST shares
## Definition of done
Our FROST implementation only accepts signed messages.

```[tasklist]
### Milestones
- [ ] Define message format for signing
- [ ] Integrate signing in coordinator
```

xoloki:
Since all `FROST` communication takes place over broadcast channels, it is necessary for `frost-signer` and `frost-coordinator` to be able to verify that messages have arrived unaltered, and come from the purported signer.  

We previously added `ECDSA` signature support to `p256k1`.  This PR uses the signature field on `FROST` network messages, and adds config support to hold private keys for the agent in question, and public keys for all `signers` and the `coordinator`.

Storing private keys this way is not good practice, but fixing that is out of scope for this PR.

All `MessageType`s are signed and verified, on both `frost-signer` and `frost-coordinator`. For now, we are calling `expect` on all `sign` operations, and `assert` for all `verify` operations.  A future PR will do more proper error handling, but this PR is big enough and for the immediate future this will suffice.  Once we have byzantine actors to worry about, and a `FIRE` or `ROAST` implementation that handles them, we will revisit this.

I also changed some of the field names in the network messages and config, to keep things rational and ease the upgrade path to `v2`.  In `v1`, a `Signer` is a set of `Party`, each of which holds one key.  In `v2`, a `Signer` is a single `Party`, each of which holds a set of keys.  So anything which is naturally described as a `key` in `v1` and `v2` is now named as such, and anything which is a `Signer` in both is named as such.  This got a little tricky in relation to `Party`, since the meaning changes in `v1` vs `v2`.  So I mostly got rid of the use of `Party`, but kept it in those places which seemed appropriate.

I also fixed a number of issues relating to networking timeouts.  Previously, we were waiting 500ms between *every* network call in `frost-signer`, and `frost-coordinator` was using an exponential backoff with a max of `3s`.  This is why it took tens of seconds for every `dkg-sign`.  I changed this so both use an exponential backoff from `2ms` to `128ms`. This is done manually in `frost-signer` and using a crate in `frost-coordinator`; given the way `frost-signer` is constructed, it would have been more trouble than it was worth to use the crate there.  

Now a `dkg-sign` round takes a few hundred milliseconds at most, even with multiple `dkg` and `nonce` rounds.

I also fixed the log spam in `HttpNetListen`.  It previously warned every time a network call failed, so if `relay-server` was not running it spammed continuously.  With the shorter timeouts this got quite bothersome.  Now it only complains the first time it fails after previously being connected.

This resolves #229


frost-coordinator/src/coordinator.rs
```diff
@@ -400,20 +424,85 @@ where
...
+                Ok(m) => {
+                    match &m.msg {
+                        MessageTypes::DkgBegin(msg) => {
+                            assert!(msg.verify(&m.sig, &coordinator_public_key))
+                        }
+                        MessageTypes::DkgPrivateBegin(msg) => {
+                            assert!(msg.verify(&m.sig, &coordinator_public_key))
+                        }
+                        MessageTypes::DkgEnd(msg) => {
+                            assert!(msg.verify(&m.sig, &signer_public_keys[msg.signer_id - 1]))
```
jferrant, line 462: I think you can condense some of these matches using something like the following (If this was done intentionally to make future changes easier, no worries..but would make it a bit easier to grok I think.):
```
MessageTypes::DkgBegin(msg) | MessageTypes::DkgPrivateBegin(msg) => {
   assert!(msg_verify(&m.sig, &coordinator_public_key))
}
```

frost-signer/src/signer.rs
```diff
@@ -1,49 +1,119 @@
...
 
 impl Signer {
-    pub fn new(config: Config, frost_id: u32) -> Self {
-        Self { config, frost_id }
+    pub fn new(config: Config, signer_id: u32) -> Self {
+        Self { config, signer_id }
     }
 
     pub fn start_p2p_sync(&mut self) -> Result<(), Error> {
+        let signer_public_keys = self
```
jferrant, line 25: This seems to come up quite a bit. Might be worth while making this into a helper function for parsing a vector public keys.

frost-coordinator/src/coordinator.rs
xoloki: Good point, that should be a lot less verbose.  I'll do that now.

frost-signer/src/signer.rs
```diff
@@ -1,49 +1,119 @@
...
             let inbound = rx.recv()?; // blocking
             let outbounds = round.process(inbound.msg)?;
             for out in outbounds {
                 let msg = Message {
-                    msg: out,
-                    sig: [0; 32],
+                    msg: out.clone(),
+                    sig: match out {
+                        MessageTypes::DkgBegin(msg) => {
+                            msg.sign(&ecdsa_private_key).expect("").to_vec()
```
jferrant, line 81: Similar to above, all of these could be combined if desired.

xoloki: I got all excited, thinking it would work for basically all the msg types that had similar fields.  Unfortunately, it only works if the actual wrapped msg type is the same :(

Fixed everything I could in both places.

jferrant: Yeah I only noticed this when I opened in visual studio and saw that they are all different msg types :( maybe in future we can combine some of them (might not need to be separate types if the wrapping in enum gives enough context but prob not. Can investigate later)

frost-signer/src/signer.rs
xoloki: I have it in both `frost-signer` and `frost-coordinator`, do you think putting it in `frost-signer/src/lib.rs` is a reasonable place?  Or should I make like a `util.rs` mod?

frost-signer/tests/net.rs
```diff
@@ -5,7 +5,7 @@ use frost_signer::signing_round::{DkgBegin, MessageTypes};
 fn receive_msg() {
     let m1 = Message {
         msg: MessageTypes::DkgBegin(DkgBegin { dkg_id: 0 }),
-        sig: [0; 32],
+        sig: [0u8; 64].to_vec(),
```
jferrant, line 8: Not sure you share Sergey's annoyance at macros, but you could replace this with the more rusty vec![0; 64] if you want to :P

stacks-signer/src/main.rs
```diff
@@ -34,10 +35,18 @@ fn main() {
...
             if let Err(e) = secp256k1.generate_private_key() {
                 warn!("An error occurred generating private key: {}", e);
             }
         }
+        Command::PublicKey { config } => match Config::from_path(&config) {
+            Ok(config) => {
+                Secp256k1::generate_public_key(&config.ecdsa_private_key);
+            }
+            Err(e) => {
+                panic!("An error occurred reading config file {}: {}", config, e);
```
jferrant, line 48: Nit: To be consistent, might want to make this a warn. Or change the warn above to a panic!

frost-signer/tests/net.rs
xoloki: Done.

frost-signer/src/signer.rs
jferrant: I think in general I prefer having util.rs files to be extra explicit. However, if you don't anticipate a lot more helper functions, not against it being in lib.rs

jferrant (approved):
Only the nits. LGTM :)

frost-signer/src/signer.rs
xoloki: Done.

stacks-signer/src/main.rs
xoloki: Yeah if main fails that's a `panic` not a `warn`.

Fixed.

Cargo.toml
```diff
@@ -25,7 +26,7 @@ rand_core = "0.6"
 hashbrown = "0.13"
 bincode = "1.3.3"
 itertools = "^0.10.5"
-sha3 = "0.10.6"
+sha2 = "0.10"
```
igorsyl, line 29: Why are we downgrading sha3 to sha2?

frost-coordinator/src/coordinator.rs
```diff
@@ -45,10 +49,17 @@ pub struct Coordinator<Network: NetListen> {
     public_nonces: BTreeMap<u32, NonceResponse>,
     signature_shares: BTreeMap<u32, v1::SignatureShare>,
     aggregate_public_key: Point,
+    ecdsa_private_key: Scalar,
```
igorsyl, line 52: Are we able to use Schnorr over ECDSA for packet signatures? If not, why not?

frost-coordinator/src/coordinator.rs
```diff
@@ -45,10 +49,17 @@ pub struct Coordinator<Network: NetListen> {
...
+    signer_public_keys: Vec<String>,
```
igorsyl, line 53: Can we create a semantic type PublicKey=String to help the reader?

frost-coordinator/src/coordinator.rs
```diff
@@ -45,10 +49,17 @@ pub struct Coordinator<Network: NetListen> {
...
+    key_public_keys: Vec<String>,
+    coordinator_public_key: String,
 }
 
 impl<Network: NetListen> Coordinator<Network> {
     pub fn new(id: usize, dkg_id: u64, config: &Config, network: Network) -> Self {
+        let ecdsa_private_key = Scalar::try_from(config.ecdsa_private_key.as_str())
```
igorsyl, line 60: Instead of calling the config `ecdsa_private_key` which encodes the sig type, can we call it something like `network_private_key` which encodes the purpose?

frost-signer/conf/signer.toml
```diff
@@ -2,6 +2,9 @@ http_relay_url = "http://localhost:9776"
 total_signers = 3
 total_keys = 6
 keys_threshold = 4
-max_party_id = 3
 frost_state_file = "frost.state.bin"
+ecdsa_private_key = "9aSCCR6eirt1NAHwJtSz4HMwBHTyMo62SyPMvVDt5DQn"
```
igorsyl, line 6: Could we also set these using environment variables?

frost-signer/src/signer.rs
```diff
@@ -72,15 +121,68 @@ impl From<mpsc::SendError<Message>> for Error {
...
+    id: u32,
+    signer_public_keys: Vec<ecdsa::PublicKey>,
+    key_public_keys: Vec<ecdsa::PublicKey>,
+    coordinator_public_key: ecdsa::PublicKey,
+) -> Result<(), Error> {
+    const BASE_TIMEOUT: u64 = 2;
+    const MAX_TIMEOUT: u64 = 128;
+    let mut timeout = BASE_TIMEOUT;
     loop {
         net.poll(id);
```
igorsyl, line 136: Does this use epoll under the hood?

frost-signer/src/signing_round.rs
```diff
@@ -105,39 +143,101 @@ pub struct DkgPublicShare {
     pub public_share: PolyCommitment,
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+impl Signable for DkgPublicShare {
+    fn hash(&self, hasher: &mut Sha256) {
+        hasher.update(self.dkg_id.to_be_bytes());
```
igorsyl, line 148: Is it common practice to add a prefix string to avoid potential sig attacks?

igorsyl (changes requested):
image

Cargo.toml
xoloki: Because we need to use sha2 for bitcoin compatibility, and if we have to use it anyway might as well use it everywhere.

frost-coordinator/src/coordinator.rs
xoloki: We certainly could use Schnorr, but it doesn't really seem to gain us anything.  We get a fast ECDSA implementation from a secp256k1, and we aren't aggregating or verifying in parallel, which are the reasons to use Schnorr.

frost-coordinator/src/coordinator.rs
xoloki: `String` is how it's stored in the config file, it's not indicative of anything semantic about the type.

frost-coordinator/src/coordinator.rs
xoloki: Sure. We could also just call it private key, it's the only private key we're storing.

frost-signer/conf/signer.toml
xoloki: We certainly could.  This PR is big enough already, so I'd prefer to save it for a follow-on.

frost-signer/src/signer.rs
xoloki: No idea what's going on under the hood.  @sergey-shandar ?

frost-signer/src/signing_round.rs
xoloki: Yeah, I can definitely add prefix strings.

xoloki: In fact, for `DkgQuery` the prefix string is the only thing we're hashing XD

frost-coordinator/src/coordinator.rs
xoloki: Done, `network_private_key` it is.

frost-signer/src/signing_round.rs
xoloki: Done.

frost-signer/src/signer.rs
xoloki: Semantically, the token `poll` here means repeatedly hitting the same endpoint until you get something you want.  Not like linux's `poll` or `epoll`, which are functionally doing a `select` not a semantic `poll`.

I really really hope the `stacks-node` version of this operates as a long poll, i.e. doesn't return until you get what you asked for.  Distributed-systems me *hates* polling :P

Cargo.toml
igorsyl: Can we use sha3 for network stuff and sha2 for bitcoin stuff?

frost-coordinator/src/coordinator.rs
igorsyl: Yes, but we can use a type alias to indicate that it's not any string but a pubkey string.

igorsyl (approved):
Approving to unblock. We should do a whole-repo review every now and then.

frost-signer/src/signing_round.rs
igorsyl: Several other hashing functions as well. Didn't tag them all.

frost-signer/conf/signer.toml
igorsyl: I'll need it for deployment. Thanks!

Cargo.toml
xoloki: Sure, let's save it for the follow on

frost-signer/src/signing_round.rs
xoloki: Refresh, latest code should have them all.

frost-signer/conf/signer.toml
xoloki: It's there, refresh XD

frost-coordinator/src/coordinator.rs
dreadful-dev: Like so:
```mermaid
graph TB
	subgraph Z["Papaya"]
		direction TB;
		C-->|DKG+Signing|B[sBTC Relay]
	    	C[Papaya coordinator]-->|Set protocol disbursement address|D[Papaya Stacking Contract]
	    	C-->|Get signer configuration|D
	    	C-->|DKG+Signing|P[Papaya Relay]
		S[Signer]-->|DKG+Signing|P
	end
	subgraph ZA["sBTC"]
		direction TB;
		G[Stacks coordinator]-->|Mint/burn sBTC|H[sBTC contract]
		G-->|Set peg wallet address|H
		G-->|Get signer configuration|H
		G-->|Get sBTC ops|N[Stacks node RPC]
		G-->|DKG+Signing|Q[Relay]
		S1[Signer]-->|DKG+Signing|Q
	end
	C-->|DKG+Signing|Q

```

PR #252, opened 2023-04-07
Setup arm build

empeje:
Fixes https://github.com/Trust-Machines/core-eng/issues/198

Prereq:

1. Owner of this repo should register to Buildjet.
2. Give access buildjet to this repo
3. merge this PR :wink: and you're good to go

The test build available in:
https://github.com/empeje/core-eng/pull/1


.github/workflows/rust.yml
```diff
@@ -29,13 +29,19 @@ jobs:
...
     runs-on: ${{matrix.os}}
     steps:
     - uses: denoland/setup-deno@v1
+      if: ${{matrix.os != 'buildjet-4vcpu-ubuntu-2204-arm'}}
       with:
         deno-version: v1.*
+    - name: Download & install deno
+      if: ${{matrix.os == 'buildjet-4vcpu-ubuntu-2204-arm'}}
+      run: |
+        curl -fsSL https://deno.land/x/install/install.sh | sh -s v1.32.3
```
sergey-shandar, line 43: Could we make it consistent and use only the `curl + install.sh` script? I mean, remove `if` and `deonland/setup-deno@1`.

.install/bitcoin.sh
```diff
@@ -3,6 +3,12 @@ then
     # URL to Bitcoin v22.0
     curl -o bitcoin.rb https://raw.githubusercontent.com/Homebrew/homebrew-core/fa6b4765d81016166f6de2bdad96cfe914c1439f/Formula/bitcoin.rb
     brew install ./bitcoin.rb
+elif [[ $(uname -m) == 'arm' ]]
+then
+    # URL to Bitcoin v22.0
+    curl -o bitcoin-22.0-arm-linux-gnueabihf.tar.gz https://bitcoincore.org/bin/bitcoin-core-22.0/bitcoin-22.0-arm-linux-gnueabihf.tar.gz
```
sergey-shandar, line 9: Can we make a variable `platform` to store either `arm-linux-gnueabihf` or `x86_64-linux-gnu`?

.github/workflows/rust.yml
empeje: I'm not 100% sure, that would work, but I'll give it a try.

empeje: I have decided that we need to use the "if" since by default the deno.land's script doesn't support aarch64 and I need to use a fork of the script that supports it. I think it is a fair compromise.

.install/bitcoin.sh
empeje: That's good idea, I have updated it using a simple map-like format.

empeje:
I have make it work and tested it in my own work where you can see in the following link:
- https://github.com/empeje/core-eng/pull/1/commits/4dd373c6722a5251d1e58d0a22bda4f159ea5af1
- https://github.com/empeje/core-eng/actions/runs/4652086719/jobs/8232128059

.install/bitcoin.sh
```diff
@@ -1,10 +1,32 @@
+val_of_platform() {
+  case $1 in
+        'darwin'*)
+            echo 'x86_64-apple-darwin'
+            ;;
+        'x86_64'*)
+          echo 'x86_64-linux-gnu'
```
netrome, line 7: Nit: Indentation mismatch

netrome (approved):
Just one nit.

.install/bitcoin.sh
empeje: Good catch, I have updated it.

stjepangolemac (approved):

PR #254, opened 2023-04-10
Broadcast bitcoin transaction to local bitcoin node

jferrant:


igorsyl (changes requested):
Should there be test cases?

jferrant:
> Should there be test cases?

See my comment in discord. I want to add integration test cases but hard to do as is.

soju-drinker (approved):
This seems good to me, I cannot see any failure cases that this PR itself is creating that can bubble up. Integration testing can be a separate PR effort

sergey-shandar (approved):
LGTM, Moving forward, I would prefer more unit tests.

PR #264, opened 2023-04-11
Set the bitcoin wallet address in the sbtc contract using aggregate public key

jferrant:
On startup of the coordinator, I force the dkg round to run and using the resulting aggregate public key, generate a bitcoin address which is then immediately used to set the sBTC contract address. 

Closes https://github.com/Trust-Machines/core-eng/issues/262


netrome (approved):
LGTM :ok_hand:

stjepangolemac (approved):

PR #268, opened 2023-04-14
Document and define commit reveal op format

### LINKED ISSUE #236: Document and define commit-reveal op format
Currently we have some scattered information on the commit-reveal operation format. We need to gather this information to a single source of truth which we can use to proceed.

## Definition of done
There is a single source of truth defining the commit-reveal format we intend to use for sBTC

netrome:
This PR introduces a design and implementation plan to deliver commit-reveal operations for sBTC.

Closes #236


sbtc-planning/commit-reveal-ops.md
```diff
@@ -0,0 +1,195 @@
...
+ magic  op
+```
+
+where the opcode is `w`, the `magic` bytes are `T2` for mainnet and `X2` for testnet.
+
+By consuming a UTXO from the _commit_ transaction, the _reveal_ ransaction will disclose
+the data embedded in the witness script. This is the data which would have been part
+of the `OP_RETURN` output in a SIP-021 operation. Anyone observing a bitcoin op with
+opcode `w` would know to look for the data in the witness of the first input, instead
+of in the rest of the `OP_RETURN`.
```
netrome, line 75: @kantai or @jcnelson I'd love to hear some feedback on this from you. This touches some fundamental logic in how Stacks nodes will parse Bitcoin transactions, and I don't want to propose anything you're not comfortable with.

From an implementation perspective, I believe this should be doable by modifying `BitcoinBlockParser::parse_tx` to look for data in witness scripts in addition to OP_RETURN data.

Do you think this makes sense? Any other ideas or proposals on the operation format? What are your thoughts?

sbtc-planning/commit-reveal-ops.md
```diff
@@ -0,0 +1,195 @@
...
+
+**6. The _Revealer_ broadcasts the _reveal_ transaction.**
+The _Revealer_ now has all the information it needs to construct and broadcast the _reveal_ transaction.
+However, the _Revealer_ is not required to do so. It may choose to ignore the commit transaction altogether,
+if it finds the cost of revealing the transaction to be too high. Therefore, the commit transaction
+output should amount to at least the output amount of the _reveal_ transaction plus the transaction fee for the miner.
+
+Security note: The _Revealer_ has the possibility to steal the commit transaction for it's own use. Doing
+this is a visible crime on the chain, but it relies on the _Revealer_ to be a trusted entity. If the _Revealer_
+is operated by Stackers, it is possible to implement mechanisms to penalize such behavior.
```
netrome, line 127: @MarvinJanssen I'd love to hear your thoughts on this, and the rest of the protocol from a security perspective. This is naturally less secure than plain OP_RETURN operations, but the question is if it's secure enough for users of custodian wallets.

sbtc-planning/commit-reveal-ops.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+## Implementation considerations
+
+### Reclaimable commit transactions
+There is no guarantee that a commit transaction will be processed by a _Revealer_. Therefore, we strongly engourage
+the Committer to include a mechanism for a user to reclaim their funds if they are not processed. This can be
+achieved in many different ways, each with their own advantages and limitations. It is up to the Committer implementation
+to design and implement this mechanism. The commit-reveal protocol is agnostic to wether or not the commit transaction
+is reclaimable.
```
netrome, line 155: @radicleart let me know if you want to add anything more specific here. Otherwise, I like to leave this open for the time being since you're driving this on your end already.

sbtc-planning/commit-reveal-ops.md
```diff
@@ -0,0 +1,195 @@
...
+of the `OP_RETURN` output in a SIP-021 operation. Anyone observing a bitcoin op with
+opcode `w` would know to look for the data in the witness of the first input, instead
+of in the rest of the `OP_RETURN`.
+
+Any remaining outputs of the _reveal_ transaction must be of the same form as in the SIP-021 format.
+For example, the _reveal_ transaction of a peg out request
+must contain two additional outputs. The first output should provide the BTC recipient address
+and the second output should fund the fulfillment transaction.
+
+## Commit-reveal protocol
```
netrome, line 82: @FriendsFerdinand I'd love to have your eyes on this, and know if this protocol lives up to your original ideas.

sergey-shandar (approved):

stjepangolemac (approved):
I'm not familiar with this enough but there's nothing obviously wrong.

sbtc-planning/commit-reveal-ops.md
```diff
@@ -0,0 +1,195 @@
...
+
+**4. The User broadcasts the commit transaction.**
+The User makes a payment to the commit address through their wallet, which results in the commit transaction
+being broadcasted to Bitcoin.
+
+**5. The _Revealer_ reads the commit transaction.**
+Having already received the witness script, and hence the address of the commit transaction, the revealer
+notices and reads the commit transaction.
+
+**6. The _Revealer_ broadcasts the _reveal_ transaction.**
```
netrome, line 119: A current problem is that there's no verification that the outputs of the reveal transaction are correct. For example, say a user want to peg-in 1 BTC. The commit transaction will encode the stacks address to receive the funds, but not the amount. The amount is determined by the first output. Hence, the _Revealer_ may choose to construct a peg-in transaction for only 0.00001 BTC and keep the rest of the funds for themselves.

One way around this problem would be to include a checksum over the transaction outputs in the OP_DROP data. We could for instance require a trailing SHA-256 hash of all outputs (value + scriptpubkey) to be added to the data. @MarvinJanssen any thoughts on this?

jferrant (approved):
Nothing stands out as a red flag to me. I imagine lots of details will start popping up during implementation.

sbtc-planning/commit-reveal-ops.md
kantai: Yeah, this definitely seems feasible from an implementation perspective.

kantai (commented):
This looks good to me, though I'm uncertain of the advantage of converting the sBTC bitcoin ops into commit/reveal pairs. If I understand correctly, the advantage is that users can peg into sBTC via the "commit" operation. The advantage of this is that the commit operation is an operation typically well supported by wallets, because it is a simple transfer to a p2tr or p2sh address. This definitely seems like an advantage to me, but are there advantages for other operations? It seems like PegOuts and PegHandoffs are less likely to be actually managed by user wallet software rather than special purpose software. If my understanding is correct, I'd suggest rephrasing this document to suggest that this is only going to be used (at first) for PegIns.

netrome:
> Sharing the gdoc that instigated commit-reveal scheme.
> 
> [sBTC proposal: Using OP_DROP instead of OP_Return](https://docs.google.com/document/d/1bbz4dY47qjZzafUFD2Pa2EaDOxDC_L1DIW4o0jo5Lkg/edit)
> 
> cc @FriendsFerdinand @tycho1212

This doc is also linked in the proposal, in the "background" section.

sbtc-planning/commit-reveal-ops.md
```diff
@@ -0,0 +1,195 @@
...
+transaction. The _Committer_ returns the commit operation address to the User, along with the witness
+script if the User wishes to verify the correctness of the address.
+
+**4. The User broadcasts the commit transaction.**
+The User makes a payment to the commit address through their wallet, which results in the commit transaction
+being broadcasted to Bitcoin.
+
+**5. The _Revealer_ reads the commit transaction.**
+Having already received the witness script, and hence the address of the commit transaction, the revealer
+notices and reads the commit transaction.
```
MarvinJanssen, line 117: I don't know if this applies to the final version of the sBTC protocol, but at least for sBTC Mini there is an additional step in that the commit transaction and its unlock script are posted to a Clarity smart contract. It allows the sBTC protocol to learn about all pending peg-out requests and also makes them visible for everyone.

sbtc-planning/commit-reveal-ops.md
MarvinJanssen: >A current problem is that there's no verification that the outputs of the reveal transaction are correct. For example, say a user want to peg-in 1 BTC. The commit transaction will encode the stacks address to receive the funds, but not the amount. The amount is determined by the first output. Hence, the Revealer may choose to construct a peg-in transaction for only 0.00001 BTC and keep the rest of the funds for themselves.

Can you explain it more? Again I am focussed on Mini but there, the user pegging in funds the commit P2TR address with the amount they want to peg-in. So the Stackers should consume the entire amount and move it to the peg wallet. The threshold signers will all independently sign a transaction consuming the output and sending it to the peg-wallet address.

I think, at least what we have been working on so far, is that there is a fourth party, the threshold-signers. In this doc it is started that the "Revealer" constructs and broadcasts the reveal transaction. By construct do you mean sign, or just assemble the transaction from the threshold signers? For Mini, anyone can submit a mined Bitcoin transaction that consumes the P2TR output to the Clarity side to prove the peg-in has happened.

sbtc-planning/commit-reveal-ops.md
MarvinJanssen: How so? If the unlock script for the peg-in is as follows then this is not a risk, unless the stackers collude.

1. Here is an 80 byte payload, but ignore it.
2. If the current block height is before N, then the threshold-signers can spend.
3. If the height is on or after N, then I can spend it.

The 80 byte payload contains the principal that should be credited the sBTC.

Once the P2TR output is spent, that Bitcoin transaction is submitted to the sBTC protocol. The Clarity contract will then determine whether it was spent by the threshold-signers or by the address in the expiry path. If the prior, mint the sBTC and credit the principal; if the latter, consider the peg-in to have failed and do nothing.

As to what is written in the doc:

>Security note: The _Revealer_ has the possibility to steal the commit transaction for it's own use.

The P2TR output is spent either by the threshold-signers or by the user in case of expiry. In that sense, the Revealer does not have the ability to steal the Bitcoin in the commit transaction. (Unless the Revealer is equal to the threshold-signers, in which case the Stackers are colluding, making sBTC  worthless.)

sbtc-planning/commit-reveal-ops.md
```diff
@@ -0,0 +1,195 @@
...
+This chart summarizes the entire commit reveal operation flow
+
+```mermaid
+sequenceDiagram
+  actor User
+  User ->> sBTC Committer: 1. Provide sBTC operation data
+  sBTC Committer -->> sBTC Revealer: 2. Send witness script and associated data
+  sBTC Committer ->> User: 3. Return commit operation address
+  User ->> Bitcoin: 4. Broadcast commit transaction
+  sBTC Revealer -->> Bitcoin: 5. Read commit transaction
```
MarvinJanssen, line 143: I assume this arrow should go from right to left `<---`.

sbtc-planning/commit-reveal-ops.md
```diff
@@ -0,0 +1,195 @@
...
+Security note: The _Revealer_ has the possibility to steal the commit transaction for it's own use. Doing
+this is a visible crime on the chain, but it relies on the _Revealer_ to be a trusted entity. If the _Revealer_
+is operated by Stackers, it is possible to implement mechanisms to penalize such behavior.
+
+**7. Stacks and sBTC read the _reveal_ transaction as any sBTC op.**
+Once the _reveal_ transaction is published, it contains the same information as a traditional `OP_RETURN` transaction
+and can be processed in the same way by the stacks node.
+
+### Summary
+This chart summarizes the entire commit reveal operation flow
```
MarvinJanssen, line 134: The design, at least for sBTC Mini is a bit different as mentioned in comments above. (The commit transactions / peg-out requests are also relayed to Stacks / the Clarity contract.)

sbtc-planning/commit-reveal-ops.md
MarvinJanssen: In sBTC Mini this was not addressed, but I added a reclaim function that can be invoked by anyone once the peg-out request has expired. (Not pegged out in 2100 blocks.) The user can also consume the P2TR output via the expiry script path to get the Bitcoin back. (But normal wallets cannot construct such a transaction?)

For a peg-in it is a lot simpler. No sBTC tokens are minted if the commit transaction is not consumed in time. The user would simply use the expiry script path to get the posted fees back. (Minus Bitcoin transaction fees.)

MarvinJanssen (commented):
I appreciate that you started on this doc. I'm adding comments keeping in mind that some things might be Mini-specific. So if some of these are redundant then please let me know.

What I would like also is to get the most up to date versions of the Taproot unlock scripts. We need those to start on the sBTC peg in/out logic. I have a reasonable idea of what they would look like but it is better to have it all laid out.

netrome:
> This looks good to me, though I'm uncertain of the advantage of converting the sBTC bitcoin ops into commit/reveal pairs. If I understand correctly, the advantage is that users can peg into sBTC via the "commit" operation. The advantage of this is that the commit operation is an operation typically well supported by wallets, because it is a simple transfer to a p2tr or p2sh address. This definitely seems like an advantage to me, but are there advantages for other operations? It seems like PegOuts and PegHandoffs are less likely to be actually managed by user wallet software rather than special purpose software. If my understanding is correct, I'd suggest rephrasing this document to suggest that this is only going to be used (at first) for PegIns.

You make a good point @kantai. While the PegIn operation is the most critical one for this protocol, I think PegOutRequests also makes sense to include in the protocol since the user must fund the transaction fee. Due to this, to request a PegOut the user must spend a small amount of their btc. If this btc is only accessible through a custodian wallet, it makes sense to support constructing the peg out request from this wallet.

I will clarify in the proposal that this will only be used at first for PegIns and PegOutRequests. However, I don't see any reason to constrain the system to these operations. If a miner would like to submit a LeaderBlockCommit with this scheme, I see no reason to explicitly prevent them from doing so. I would however be very surprised to see that happen.

sbtc-planning/commit-reveal-ops.md
netrome: Nice, thank you for confirming! I interpret your comment as:

1. It seems possible to implement it
2. We won't totally wreck the Stacks code base/architecture by implementing this

I assume more considerations and concerns are likely to pop up as we begin implementing this, but it seems like we can start the process of making this real soon.

sbtc-planning/commit-reveal-ops.md
netrome: Sure, in our commit-reveal implementation the Revealer will be the threshold-signers so it should be both difficult for them to collude and if they do so we are able to freeze their stacked STX and repurpose their PoX rewards to compensate any victim in the spirit of the SIP-021 recovery mode.

sbtc-planning/commit-reveal-ops.md
netrome: Alright. I think the peg-in is the most critical transaction. The peg-out request is a very small transaction, since it is only used to fund a transaction fee. I'm not sure if reclaiming such a transaction even makes sense, since the reclaim transaction fee would eat up most or all of the funds.

sbtc-planning/commit-reveal-ops.md
netrome: Yeah, in this context I would view that as an implementation detail of the sBTC mini version of this protocol. This is essentially a way for the Committer (sBTC Bridge) to communicate with the Revealer (threshold signers) in a very visible and transparent way. Does that make sense to you?

sbtc-planning/commit-reveal-ops.md
radicleart: agreed. its the peg-in that needs the reclaim. I think we got to that this has to be a CSV custom script since a multisig script of any kind breaks the op_drop tx mechanism.

sbtc-planning/commit-reveal-ops.md
netrome: Yeah I commented in another thread; I view sBTC Mini as a concrete implementation of this protocol. The fact that sBTC Mini uses a Clarity contract is a nice feature for transparency, and does not conflict with anything specified here afaik. I'd like to keep the base protocol as raw and simple as possible. I hope this makes sense to you, and I'm happy to further explore the sBTC Mini needs with you if it doesn't.

radicleart:
> It seems like PegOuts and PegHandoffs are less likely to be actually managed by user wallet software rather than special purpose software. If my understanding is correct, I'd suggest rephrasing this document to suggest that this is only going to be used (at first) for PegIns.

Assuming there are no technical reasons why not, it makes sense from a UX point of view for both the user initiated bitcoin transactions (PegIn Commit and PegOut Request) to use op_drop mechanism from the get go. Otherwise we'd be creating a barrier to them being able to pegout once they have pegged in - e.g. from their Coinbase account.

radicleart:
@netrome thanks, this is a really helpful document!

When your'e ready for it the witness script and associated data for the reveal transactions (simple path with no reclaim) is being stored via the api and can be redirected any time.

netrome:
> @netrome thanks, this is a really helpful document!
> 
> When your'e ready for it the witness script and associated data for the reveal transactions (simple path with no reclaim) is being stored via the api and can be redirected any time.

Happy it's helpful. I guess @MarvinJanssen is interested in the details of those scripts as you are working on them, so please sync this with each other as it evolves. Just a head's up. I'm extending the proposal to add a checksum to the OP_DROP data to allow reveal outputs to be validated against the commit data.

netrome:
@kantai @MarvinJanssen @radicleart - thank you very much for your feedback. I added two changes based on the discussion:

1. Introduce a sha256 checksum of outputs in the data.
2. Clarify in the implementation considerations that we initially will focus on supporting PegIn and PegOutRequest.

I believe all major known concerns are addressed in the current version. Let me know if you have any further considerations or reservations. Otherwise, my intention is to begin implementing this in Stacks and drafting a SIP during next week.

netrome:
After a discussion in todays sBTC Eng meet, I have two changes pending:

1. Rename `checksum` to `output commitment hash`
2. Reduce the size of the checksum to save storage space.

On 2. I'll update the requirement to be 12 bytes. With the current bitcoin hash rate of 330.89M TH/s [source](https://ycharts.com/indicators/bitcoin_network_hash_rate), this would take $\frac{2^{8*12}}{330.89*10^12} \text{s} \approx 239 \text{Ts} \approx 7 \text{million years} $

sbtc-planning/commit-reveal-ops.md
MarvinJanssen: Ok it became more clear during our meeting.

>Technically, the revealer (i.e. the signers) could pay almost the entire committed amount as a giant transaction fee and only peg in a dust amount.
>
>What I'm thinking here is that if we include a checksum over the outputs (amount + scriptPubKey) in the committed payload we should solve this problem, because it would make any malicious activity by the revealer explicit.

I'm not sure if there is a lot of upside to a checksum. We should ask ourselves what problem is really being solved. In the end the Stackers can spend the entire output regardless of what data is stuffed in there. So if they decide to seize the output for themselves then the protocol is dead anyway.

If the fee is just a uint in the witness script then it would be much easier for Clarity to interpret it and check it. Then it is also much simpler to prove to a Clarity contract that the Stackers took too large a fee and credit the user back some extra sBTC or something. If the user says "I'm willing to pay X fees for a peg-in" then it is up to the Stackers whether or not to process the peg-in and to honour that number.

Fees have not really been figured out anyway as far as I understand. In sBTC Mini it was argued that the fee should be fixed but I do not think it is workable nor desirable. In any case, I assume that whatever fee mechanism will exist, that the fee logic will be added to a Clarity contract.

MarvinJanssen:
I added a comment regarding the checksum / hash. I am not sure it is necessary for the mechanism described. Also, why are you concerned about how long it takes to find a collision?

sbtc-planning/commit-reveal-ops.md
netrome: You're right. I was working under the assumption that there was more information than the fee which we must commit to, since the peg out request contains the peg out address in its output as well. However, this address and the peg out amount is actually committed to in the peg out data with a signature - so the only information we need to add is the fee for the PegIn.

I'll change the proposal to just require adding the fee as "I am willing to pay up to X in fees for the reveal transaction".

netrome:
> I added a comment regarding the checksum / hash. I am not sure it is necessary for the mechanism described. Also, why are you concerned about how long it takes to find a collision?

Replied to your comment, you have convinced me and I will update the proposal to not include a checksum.

The time it takes to find a collision would be the time it would take for a malicious actor to find another valid output which the user did not commit to, allowing the actor to burn (this address is most likely unspendable) the funds if it wanted to. But yeah nevermind, let's kill the checksum with fire :fire:

netrome:
I've updated now and removed the checksum in favor of simply specifying the maximum allowed reveal transaction fee. Thank you @MarvinJanssen for questioning the checksum. This is a much simpler solution.

netrome:
I'm very satisfied with the current version, and I will merge this design now. This does not mean the design is finalized. We will iterate on this as we learn.

PR #282, opened 2023-04-21
Remove unnecesary comment, body already sent into the req

stjepangolemac:
As the title says, just a one line change to clean up the transaction broadcasting (which we suspect is already set up correctly so no additional corrections should be needed).


netrome (approved):
I have vetted this thoroughly and not been able to find a single flaw in neither design nor execution of this well-crafted piece of engineering.

stjepangolemac:
Closing as the branch was renamed.

PR #285, opened 2023-04-22
Load wallet into bitcoind to retrieve UTXOs

jferrant:
This PR is for adding in missing pieces to integrate with bitcoind. (was not previously actually loading the bitcoin wallet so calls to listunspent would never work). 

Closes https://github.com/Trust-Machines/core-eng/issues/298


stacks-coordinator/src/bitcoin_node.rs
```diff
@@ -97,18 +104,23 @@ impl LocalhostBitcoinNode {
...
                 let err_str = e.to_string();
-                let error_json = serde_json::json!({ "error": &err_str });
-                let err_obj_opt = match e.into_response() {
-                    Some(r) => r.into_json::<serde_json::Value>().unwrap_or(error_json),
-                    None => error_json,
+                let error_json = serde_json::json!({ "error": {"message": &err_str }});
+                let error_message = match e.into_response() {
+                    Some(r) => {
+                        let json = r.into_json::<serde_json::Value>().unwrap_or(error_json);
+                        let message = &json["error"]["message"];
```
netrome, line 118: This may panic. I recommend using the `.get` method instead and handling potential errors..

jferrant: Good catch. This was a placeholder I had resorted to because its inside a mapped err which cannot handle returning a Result so would need to explicitly call unwrap. Mostly this is done so I can do some proper testing, however it definitely should be merged as is and I will add an explicit TODO to make sure it is not.

netrome (approved):

stjepangolemac (approved):

PR #289, opened 2023-04-24
Create AI_POLICY.md

igorsyl:


netrome:
Since this is just an AI policy and not contribution guidelines, I suggest renaming the file to `AI_POLICY.md`. Personally I don't see the need to explicitly not prohibit something, but Igor felt strongly for this.

netrome (approved):
Sure, why not.

stjepangolemac (approved):

friedger:
This implies that contributions to this repo should not be taken into account for reputation, coding skills etc of the submitter themselves.

friedger:
I assume that only good code is submitted and accepted.

If we can't provide tools and evidence that the code was hand crafted then we can assume that all code was generated.

netrome:
> I assume that only good code is submitted and accepted.
> 
> If we can't provide tools and evidence that the code was hand crafted then we can assume that all code was generated.

I wouldn't make that assumption, but sure. In the absence of such tools/evidence we can't know how any code (except our own) has been crafted.

soju-drinker:
> I assume that only good code is submitted and accepted.
> 
> If we can't provide tools and evidence that the code was hand crafted then we can assume that all code was generated.

What exactly is the issue with accepting AI generated code that is high quality and passes both the owner and the reviewer's eyes?

PR #291, opened 2023-04-25
Relay server improvements from an old PR #50

sergey-shandar:
The PR contains only relay-server API improvements from an old PR #50


yarpc/src/http/method.rs
```diff
@@ -0,0 +1,42 @@
...
+const GET: &str = "GET";
+const POST: &str = "POST";
+
+impl Method {
+    pub const fn to_str(self) -> &'static str {
+        match self {
+            Method::GET => GET,
+            Method::POST => POST,
+        }
+    }
```
netrome, line 20: This method does not need to take ownership of self
```suggestion
    pub const fn as_str(&self) -> &'static str {
        match self {
            Method::GET => GET,
            Method::POST => POST,
        }
    }
```

netrome: In addition, I suggest implementing `Display` for this type.

yarpc/src/http/method.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub fn try_parse(value: &str) -> Option<Self> {
+        match value {
+            GET => Some(Self::GET),
+            POST => Some(Self::POST),
+            _ => None,
+        }
+    }
```
netrome, line 27: I suggest using the [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) trait here. This would make the code more idiomatic, and allow us to call "get".parse()? to parse strings.

xoloki (approved):
LGTM

yarpc/src/http/method.rs
```diff
@@ -0,0 +1,42 @@
...
+    pub fn request(
+        self,
+        url: String,
+        headers: HashMap<String, String>,
+        content: Vec<u8>,
+    ) -> Request {
```
netrome, line 33: Opinion: This should be a constructor on the `Request` type imo. I'd argue that
```
Request::new(method, url, headers, content)
```
reads better than
```
Method::POST.request(url, headers, content)
```

netrome (approved):
Some comments to make the code more idiomatic, but nothing major.

yarpc/src/http/request.rs
```diff
@@ -49,7 +34,7 @@ impl Message for Request {
 
     fn first_line(&self) -> Vec<String> {
         [
-            self.method.to_owned(),
+            self.method.to_str().to_string(),
```
netrome, line 37:
```suggestion
            self.method.to_string(),
```
(This assumes Display being implemented on Method)

relay-server/src/proxy_state.rs
```diff
@@ -0,0 +1,80 @@
...
+mod tests {
+    use super::super::*;
+    use super::*;
+
+    #[test]
+    fn test() {
+        let mut state = ProxyState(Server::default());
+        assert!(state.get(1.to_string()).unwrap().is_empty());
+        assert!(state.get(3.to_string()).unwrap().is_empty());
+        // assert_eq!(0, state.highwaters.len());
```
stjepangolemac, line 38: I'd suggest either removing or leaving a TODO.

yarpc/src/http/method.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl Method {
+    pub const fn to_str(self) -> &'static str {
+        match self {
+            Method::GET => GET,
+            Method::POST => POST,
+        }
+    }
+    pub fn try_parse(value: &str) -> Option<Self> {
+        match value {
```
stjepangolemac, line 22: Maybe we want to do `to_uppercase()` here?

stjepangolemac (approved):
A few nitpicks but otherwise good 👍

yarpc/src/http/method.rs
sergey-shandar: No. HTTP methods are case-sensitive https://www.rfc-editor.org/rfc/rfc9110.html#section-9.1-5. But I will add a comment.

yarpc/src/http/method.rs
sergey-shandar: I like the pipe operator semantic (which is `.` in Rust) so I'm trying to use it everywhere, but I agree that in this case it doesn't make much sense. The pipe operator semantic is useful when you would like to avoid nested expressions, instead of `f(g(x), y)` we would like to use this `g(x).f(y)`. But, I would expect that `method` is not a complex expression usually, so it's ok to switch to `new`.

yarpc/src/http/method.rs
sergey-shandar: This enum is cloneable, copyable ant it's better to pass by value (like u8).

netrome: Fair enough.

yarpc/src/http/method.rs
netrome: Good point. I will consider applying the pipe semantics more in my own code.

PR #301, opened 2023-04-27
295 stacks coordinator write tests to spin up bitcoind and test all bitcoin node calls

stjepangolemac:
I've stubbed out the test-utils and made them used in frost-test and should be used for stacks-coordinator integration tests.


stjepangolemac:
@jferrant I've set up the testing environment but need help with a couple of things.

- How do I properly configure the stacks-coordinator?
- Can you give me an example function that you would test in an integration set, or do we want to test it completely by sending the commands to it in the test environment?

jferrant:
> @jferrant I've set up the testing environment but need help with a couple of things.
> 
>     * How do I properly configure the stacks-coordinator?
> 
>     * Can you give me an example function that you would test in an integration set, or do we want to test it completely by sending the commands to it in the test environment?

I would probably not bother with the stacks-coordinator running entirely just yet. Sayak is working on the stacks node side. This is more to spin up a LocalHostBitcoinNode::new(bitcoind_api) and then testing all the functions such as broadcast_transaction, load_wallet, list_unspent, etc.

test-utils/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+        let mut url: Url = BITCOIND_URL.parse().unwrap();
+        url.set_port(Self::find_port()).unwrap();
+
+        let child = Self::spawn(url.port().unwrap());
+
+        let this = Self { url, child };
+        this.connectivity_check().unwrap();
+
+        this
+    }
```
netrome, line 111: Can we fix so we can run multiple BitcoinProcess instances in parallel? Currently the tests fail when I run `cargo nextest run` because of this call panicking.

stjepangolemac: Will look into that, I changed the API a bit but it still works the same way as before. Will investigate.

stjepangolemac: They pass for me, what's the error?

netrome: Here's the error I get
```
$ cargo nextest run
...

        PASS [   0.864s] stacks-coordinator::btc_node_calls btc_node_calls
        FAIL [   0.864s]              frost-test::frost-btc frost_btc

--- STDOUT:                           frost-test::frost-btc frost_btc ---

running 1 test
{"id":"tst","jsonrpc":"1.0","method":"uptime","params":null} -> {"error":"http://abcd:abcd@localhost:18443/: Connection Failed: Connect error: Connection refused (os error 111)"}
user private key cVqvByc1XdKut9t4EkNLQJeqcjq8jHAyoLfm45cRRvJP1xTzky2c
user address bcrt1qnw0kk3lw4mukjctvu0d2g5fnlck8ryy2h0l6c8 public key 03772773439838b366b7c9369aadac60b5c6194325d594e997a5181bdedf1263d7 witness hash 9b9f6b47eeaef969616ce3daa45133fe2c71908a p2wpkh signing script Script(OP_DUP OP_HASH160 OP_PUSHBYTES_20 9b9f6b47eeaef969616ce3daa45133fe2c71908a OP_EQUALVERIFY OP_CHECKSIG)
{"id":"tst","jsonrpc":"1.0","method":"generatetoaddress","params":[128,"bcrt1qnw0kk3lw4mukjctvu0d2g5fnlck8ryy2h0l6c8"]} -> {"error":"http://abcd:abcd@localhost:18443/: Network Error: Unexpected EOF"}
test frost_btc ... FAILED

failures:

failures:
    frost_btc

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.83s


--- STDERR:                           frost-test::frost-btc frost_btc ---
Error: Cannot obtain a lock on data directory /home/marten/.bitcoin/regtest. Bitcoin Core is probably already running.
thread 'frost_btc' panicked at 'called `Option::unwrap()` on a `None` value', frost-test/tests/frost-btc.rs:157:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

   Canceling due to test failure: 0 tests still running
------------
     Summary [   0.865s] 51 tests run: 50 passed, 1 failed, 3 skipped
        FAIL [   0.864s]              frost-test::frost-btc frost_btc
error: test run failed
```

stjepangolemac: @netrome try with the latest code and see if it still happens.

stjepangolemac:
I have a problem properly testing the `list_unspent()` function as bitcoind always returns an empty array as the result, even after I mine some BTC to an address.

Let me know if you have any ideas on what to check or change. cc @jferrant @netrome @sergey-shandar

netrome:
> I have a problem properly testing the `list_unspent()` function as bitcoind always returns an empty array as the result, even after I mine some BTC to an address.
> 
> Let me know if you have any ideas on what to check or change. cc @jferrant @netrome @sergey-shandar

How long are you waiting before calling listunspent? The default is to require 1 confirmation [see docs](https://developer.bitcoin.org/reference/rpc/listunspent.html#argument-1-minconf) which I would assume meaning you have to mine at least one additional block before the UTXO is visible as unspent in the call.

stjepangolemac:
When I try to mine more blocks I get this error:

```
[stacks-coordinator/src/bitcoin_node.rs:108] &e = Error("invalid number", line: 0, column: 0)
thread 'should_list_unspent' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidResponseJSON("invalid number")', stacks-coordinator/tests/local_bitcoin_node.rs:128:55
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It's coming from the bitcoind process I believe but I couldn't find much on the internet.

netrome:
> When I try to mine more blocks I get this error:
> 
> ```
> [stacks-coordinator/src/bitcoin_node.rs:108] &e = Error("invalid number", line: 0, column: 0)
> thread 'should_list_unspent' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidResponseJSON("invalid number")', stacks-coordinator/tests/local_bitcoin_node.rs:128:55
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> It's coming from the bitcoind process I believe but I couldn't find much on the internet.

This is a deserialization issue. Our current `UTXO` type is not compatible with the returned result from the listunspent call. Here is a struct which successfully deserializes:

```
#[allow(non_snake_case)]
#[derive(Debug, Deserialize, Serialize, Default)]
pub struct UTXO2 {
    pub txid: String,
    pub vout: u32,
    pub address: String,
    pub label: String,
    pub scriptPubKey: String,
    pub amount: f64, // Changed to float
    pub confirmations: u64,
    //pub redeemScript: String,
    //pub witnessScript: String,
    pub spendable: bool,
    pub solvable: bool,
    //pub reused: bool,
    //pub desc: String,
    pub safe: bool,
}
```

This could be used as an intermediary type to construct the internal UTXOs by populating some missing fields and converting the amount to an integer.

jferrant:
This is so good to see. Really appreciate this work @stjepangolemac !

netrome (approved):
Good stuff! Sorry for taking so long with the review :pray:

stacks-coordinator/src/bitcoin_node.rs
```diff
@@ -112,7 +112,13 @@ impl BitcoinNode for LocalhostBitcoinNode {
                 "Listunspent response is not an array".to_string(),
             ))?
             .iter()
-            .map(Self::raw_to_utxo)
+            .map(|val| {
+                let res = Self::raw_to_utxo(val);
+
+                dbg!(&val, &res);
```
jferrant, line 118: I would remove this dbg statement.

PR #304, opened 2023-04-27
Clarity/clarity bitcoin segwit

FriendsFerdinand:
adding segwit parsing to the clarity-bitcoin contract

- [x] Handle Coinbase TX with multiple outputs with the `0x6a24aa21a9ed` prefix

Possible features:
- [ ] Potentially handle empty Bitcoin block and the coinbase transaction is the segwit transaction we're interested in?


sbtc-mini/tests/clarity-bitcoin_test.clar
MarvinJanssen, line 1: Could you provide a human-readable representation of these test vectors?

sbtc-mini/contracts/clarity-bitcoin.clar
```diff
@@ -1,18 +1,483 @@
...
 	(tx (buff 4096)) ;; tx to check
 	(header (buff 80)) ;; bitcoin block header
 	(tx-index uint)
 	(tree-depth uint)
 	(wproof (list 14 (buff 32))) ;; merkle proof for wtxids
-	(ctx (buff 1024)) ;; coinbase tx, contains the witness root hash
+  (witness-merkle-root (buff 32)) ;; merkle root of wtxids
+  (witness-reserved-data (buff 32)) ;; merkle root of wtxids
+	(ctx (buff 1024)) ;; non-segwit coinbase tx, contains the witness root hash
 	(cproof (list 14 (buff 32))) ;; merkle proof for coinbase tx
```
MarvinJanssen, line 450: Why were these two parameters added? Can it not be extracted from the Coinbase tx `wtx`?

MarvinJanssen (changes requested):
Awesome! Nice .clar test file too.

sbtc-mini/contracts/clarity-bitcoin.clar
FriendsFerdinand: No because they are hashed before being inserted in the commitment output (see line 459)

sbtc-mini/contracts/clarity-bitcoin.clar
```diff
@@ -1,18 +1,507 @@
...
+;; Verify that a block header hashes to a burnchain header hash at a given height.
+;; Returns true if so; false if not.
+(define-read-only (verify-block-header (headerbuff (buff 80)) (expected-block-height uint))
+    (match (get-bc-h-hash expected-block-height)
+        bhh (is-eq bhh (reverse-buff32 (sha256 (sha256 headerbuff))))
+        false))
+
+(define-read-only (verify-block-header-test (headerbuff (buff 80)) (expected-block-height uint))
+  true
+)
```
MarvinJanssen, line 355: We should create a mock burn block header registry so that we can make tests more true to life.

Something like:

```clarity
(define-map mock-burnchain-header-hashes uint (buff 32))

(define-public (mock-add-burnchain-block-header-hash (burn-height uint) (hash (buff 32)))
 (ok (map-set mock-burnchain-header-hashes burn-height hash))
)

(define-read-only (get-bc-h-hash (bh uint))
   (if DEBUG-MODE (map-get? mock-burnchain-header-hashes bh) (get-burn-block-info? header-hash bh))
```

jo-tm:
@FriendsFerdinand marvin requested some changes on this PR, please merge with or without changes, so we can have smaller PRs in time and #commits.

FriendsFerdinand:
@jo-tm review is required.

sbtc-mini/contracts/clarity-bitcoin.clar
```diff
@@ -1,18 +1,518 @@
...
+      )
+    error (err u1)
+  )
+)
+
+(define-read-only (read-witnesses (ctx { txbuff: (buff 1024), index: uint }) (num-txins uint))
+  (fold read-next-witness (unwrap-panic (slice? (list true true true true true true true true) u0 num-txins)) (ok { ctx: ctx, witnesses: (list) }))
+)
+
+;; Helper functions for smart contract that want to use information of a Bitcoin transaction
```
jo-tm, line 259: Can we remove this commented code @FriendsFerdinand ? Maybe you can move it into an issue on backlog?

jo-tm (changes requested):
Requested minor changes on commented code.

sbtc-mini/contracts/clarity-bitcoin.clar
FriendsFerdinand: It's not commented code, it describes the function's output. You raised a good point and I forgot to update this to reflect segwit transactions.

jo-tm: excellent

jo-tm:
approved on my side, @friedger @MarvinJanssen  @setzeus  can you please review or comment.

sbtc-mini/contracts/clarity-bitcoin.clar
```diff
@@ -8,11 +462,62 @@
 	(tx-index uint)
 	(tree-depth uint)
 	(wproof (list 14 (buff 32))) ;; merkle proof for wtxids
-	(ctx (buff 1024)) ;; coinbase tx, contains the witness root hash
+  (witness-merkle-root (buff 32)) ;; merkle root of wtxids
+  (witness-reserved-data (buff 32)) ;; merkle root of wtxids
+	(ctx (buff 1024)) ;; non-segwit coinbase tx, contains the witness root hash
 	(cproof (list 14 (buff 32))) ;; merkle proof for coinbase tx
```
MarvinJanssen, line 468: I guess for now my only comment is what I mentioned on the call. We can get rid of the list because the Coinbase transaction is always the first, so we only need to provide the right leaf hash.

MarvinJanssen (approved):
Looks good to me. There are various points where we can do runtime cost optimisations, but I think that is best done in future PRs. With this merged we have a base that we can build on.

sbtc-mini/contracts/clarity-bitcoin.clar
FriendsFerdinand: We can't get rid of the list, we still need to hash our way to the leaf at position 0:
```
                   [Root]
                    /    \
             [Hash A]   [Hash B]
              /   \        /   \
       [Hash A1] [Hash A2] [Hash B1] [Hash B2]
       /  \        /  \          /   \         /  \
    [L1] [L2]    [L3] [L4]     [L5] [L6]      [L7] [L8]
```

(B, A2, L2)

so we know we hash left left left, but that's all we can optimize for

MarvinJanssen: Ah you're right what was I thinking. First in the list obviously does not mean the first leaf.

PR #307, opened 2023-04-28
Rust code coverage in CI

sergey-shandar:
Rust code coverage.


.github/workflows/rust.yml
```diff
@@ -62,6 +62,43 @@ jobs:
     - name: Run tests
       run: cargo test --verbose --locked ${{matrix.type}}
 
+  coverage:
+    needs: build
+    runs-on: ubuntu-latest
```
netrome, line 67: Opinion: We should use explicit versions for the docker images.
```suggestion
    runs-on: ubuntu-22.04
```

netrome (approved):
Just one suggestion.

jferrant (approved):

.github/workflows/rust.yml
sergey-shandar: I would prefer not to specify a specific version but rather make sure that it works on the latest. Similar strategy we use for `Cargo.lock`. For reproducibility, we publish the `Cargo.lock` file and GitHub CI prints the OS version, so we can reproduce a specific build if needed. https://github.com/Trust-Machines/core-eng/actions/runs/4831848681/jobs/8609839003?pr=307#step:1:2

netrome: My problem with this strategy in CI is that if a new Ubuntu image is published and our actions are not compatible with that one, then every open PR will be blocked. I find it more productive to have our CI working as deterministically as possible. 

We experienced an instance of this problem in Stacks in December when suddenly the CI failed on all open PRs, due to CI updating the deno image for some Clarity tests. I had to push a separate PR to fix this to unblock my work at the time, which was both time consuming and frustrating.

Indeed, it was easy to unblock CI but I would argue that it's much less annoying to make these things stable by default rather than always working under the risk of sudden unrelated CI failures. I don't see a big value in making sure that things always work on latest. If our builds would happen to not be compatible with the latest and greatest ubuntu version, I would not consider that an urgent problem requiring us to block everything else we're working on.

Regarding, `Cargo.lock` I see we use the `--locked` argument in our actions ensuring we use the locked version and not the latest version of the packages in CI. This aligns more with explicitly setting the docker image version I would say. If we want to make a package management analogy, I would compare this with setting wildcard versions in `Cargo.toml` (i.e. `serde = "*"`). This also prioritizes working on the latest versions of everything, but sacrifices stability.

Pheew, that was longer than I intended and I realized I feel relatively strong for this. To summarize my main point I would say:
- Stability in CI is more important than ensuring compatibility with latest

Let me know what you think about this point.

sergey-shandar: I propose to move this discussion to Discussion #309  :-)

netrome: Sounds good. Let's continue the discussion there :+1:

PR #311, opened 2023-05-01
Update sBTC contract to use signer instead of party, and put key-ids into map

xoloki:


sbtc-ops/clarinet/contracts/sbtc-alpha.clar
```diff
@@ -24,14 +24,14 @@
...
-(define-data-var num-parties uint u4000)
+(define-data-var num-signers uint u4000)
 (define-data-var threshold uint u2800)
 (define-data-var bitcoin-wallet-address (optional (string-ascii 72)) none)
 (define-data-var trading-halted bool false)
 
 ;; data maps
 ;;
-(define-map signers uint {addr: principal, key: (buff 33)})
+(define-map signers uint {addr: principal, key: (buff 33), ids: (list 4000 uint)})
```
jferrant, line 34: Would a user then use get-signer-data again on the list of frost-ids?

xoloki: No.  `get-signer-data` now actually returns data for a signer not a key.  so you would iterate over the interval `[1, get-num-signers]`, call `get-signer-data` for each of those.  Then in frost config you set `signer_public_keys` from this data.

If we want to keep the current frost config as is, you would also iterate over the returned `ids` and set `key_public_keys` for these IDs to be the public key from the `signer` who owns those keys.

We could instead change the frost config to actually keep a map of which IDs belong to which signer, but we'd also need a reverse map because some messages come in with key_id only, so we'd need to reverse lookup to find out which signer that key belongs to, and get the pubkey from there.

Does that make sense?

jferrant: No I am not advocating for the current config. Just not understanding how a list of frost ids is used.

jo-tm (approved):

jo-tm:
Maybe we can add more test cases for abnormal cases?

sbtc-ops/clarinet/contracts/sbtc-alpha.clar
xoloki: This conversation is listed as unresolved, are you good after our zoom yesterday @jferrant ?

PR #318, opened 2023-05-03
Update config to better match sBTC contract changes

jferrant:
250 lines code....500 lines tests...
Closes https://github.com/Trust-Machines/core-eng/issues/312


xoloki (approved):
Couple of quick questions and minor nits, but looks great!

frost-signer/src/signer.rs
```diff
@@ -146,43 +137,137 @@ fn poll_loop(
...
+
+fn verify_msg(
+    m: &Message,
+    signer_keys: &SignerKeys,
+    coordinator_public_key: &ecdsa::PublicKey,
+) -> bool {
+    match &m.msg {
+        MessageTypes::DkgBegin(msg) | MessageTypes::DkgPrivateBegin(msg) => {
+            if !msg.verify(&m.sig, coordinator_public_key) {
+                tracing::warn!("Received a DkgPrivateBegin message with an invalid signature.");
```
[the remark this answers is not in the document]
jferrant: Done!

frost-signer/src/config.rs
```diff
@@ -41,6 +56,34 @@ impl Config {
...
+            let signer_public_key = parse_public_key(&s.public_key).map_err(|_| {
+                Error::InvalidPublicKey(format!(
+                    "Failed to parse signers from config. {}",
+                    s.public_key
+                ))
+            })?;
+            s.key_ids.iter().for_each(|key_id| {
+                signer_keys.key_ids.insert(*key_id, signer_public_key);
+            });
+            // We start our signer ids from 1 not 0, hence i + 1
```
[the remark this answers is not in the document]
jferrant: I enforce this now when creating a Config, however, should we? I mean I was just following what was previously enforced. However, don't see why this is necessary...?

frost-coordinator/src/coordinator.rs
```diff
@@ -68,20 +65,16 @@ pub struct Coordinator<Network: NetListen> {
...
+    pub fn new(id: u32, dkg_id: u64, config: &Config, network: Network) -> Result<Self, Error> {
         let network_private_key = Scalar::try_from(config.network_private_key.as_str())
             .expect("failed to parse network_private_key from config");
 
-        Self {
-            id: id as u32,
+        Ok(Self {
+            id,
             current_dkg_id: dkg_id,
-            current_dkg_public_id: 1,
```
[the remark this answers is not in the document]
xoloki: dkg id should refer to dkg as a whole, so it should inc when dkg starts.  dkg public id should start at 0 when a new dkg is started, and inc every time we do a new public round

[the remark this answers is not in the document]
jferrant: Done :)

frost-coordinator/src/coordinator.rs
```diff
@@ -212,7 +201,7 @@ where
                 }
             }
 
-            if self.public_nonces.len() == self.total_signers {
+            if self.public_nonces.len() == usize::try_from(self.total_signers).unwrap() {
```
stjepangolemac, line 204: Seems like total signers is u32 and therefore it can always be casted to an usize. This would be simpler:

```
self.total_signers as usize
```

stjepangolemac (approved):

frost-coordinator/src/coordinator.rs
xoloki: I guess on a 16-bit platform where we had more than 65K signers, it could fail :P

netrome: Well technically speaking this could break on 16 bit platforms. `usize` is only infallibly convertible from `u16` and below. Although the `unwrap()` completely defeats the purpose of using a fallible conversion here :sweat_smile: I'd prefer it if the error was explicitly handled, but I'm also fine with the `as` conversion.

image

xoloki: Does an `as` cast `panic` if it fails @netrome ?  If so then we can just stick with that.

xoloki: So the explicit conversion plus `unwrap` seems like the safest bet, though obviously it's a really really unlikely occurrence.  But we are writing crypto code here so I say `panic`

frost-signer/src/config.rs
```diff
@@ -36,17 +40,160 @@ pub struct Cli {
...
+    pub fn from_path(path: impl AsRef<std::path::Path>) -> Result<RawConfig, Error> {
         let content = fs::read_to_string(path)?;
         Ok(toml::from_str(&content)?)
     }
+
+    pub fn signer_keys(&self) -> Result<SignerKeys, Error> {
+        let mut signer_keys = SignerKeys::default();
+        for (i, s) in self.signers.iter().enumerate() {
+            let signer_public_key = parse_public_key(&s.public_key).map_err(|_| {
+                Error::InvalidPublicKey(format!(
```
netrome, line 69: Nit: I think it would be nice here if `Error::InvalidPublicKey` would wrap the error returned from `parse_public_key` instead of stringifying it. If we also implement From<ParsePublicKeyError> for the Error type, you could just use the `?` operator here and save us some boilerplate.

netrome (approved):
Just a bunch of nits and smaller comments. Otherwise LGTM.

frost-coordinator/src/coordinator.rs
netrome: Interesting, I was under the impression that `as` would panic.

netrome: Verified this on my machine as well. Then yeah I agree the explicit unwrap is better than the `as` conversion.

frost-signer/src/config.rs
netrome: Ah I see. Yeah wrapping is not worth it if we sacrifice the ability to derive `Debug` or `Display` (although these could of course be explicitly defined).

I see the base58 error type here does not implement `Display`: https://docs.rs/base58/0.2.0/base58/enum.FromBase58Error.html. That's weird. All decent error types should be compatible with `std::error::Error` https://doc.rust-lang.org/std/error/trait.Error.html.

netrome: I created an issue in the repo, but it does not seem to be under active development https://github.com/debris/base58/issues/5

netrome: Another issue, I don't see a license in the repo although it's marked as MIT licensed in `crates.io` :thinking: https://github.com/debris/base58

netrome: This base58 library seems better maintained. We should consider switching to it https://crates.io/crates/bs58

netrome: I created an issue to switch. Let me know if you have any opinions: https://github.com/Trust-Machines/core-eng/issues/320

frost-coordinator/src/coordinator.rs
jferrant: Yeah its a recommended practice (though it irks my soul in general to have unwraps in code...this is the only place I suck it up and just do..) to force a panic if the conversion fails otherwise you can end up with bugs popping up elsewhere that are hard to debug. Basically results in undefined behaviour if the "as" conversion fails/

frost-signer/src/config.rs
jferrant: Yeah this is why i didn't do, unfortunately. :(

frost-signer/src/config.rs
```diff
@@ -36,17 +40,160 @@ pub struct Cli {
...
+mod test {
+    use crate::config::{RawConfig, RawSignerKeys};
+    #[test]
+    fn coordinator_public_key_test() {
+        let mut config = RawConfig::default();
+        // Should fail with an empty public key
+        assert!(config.coordinator_public_key().is_err());
+        // Should fail with an invalid public key
+        config.coordinator_public_key = "Invalid Public Key".to_string();
+        assert!(config.coordinator_public_key().is_err());
```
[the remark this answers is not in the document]
jferrant: OOo I had wanted to do this but it didn't implement partial eq and couldn't because of certain error types. Didn't know about the macro "matches!". Thanks!

PR #326, opened 2023-05-08
Change get-bitcoin-wallet-address to get-bitcoin-wallet-public-key

jferrant:
We need the bitcoin public key in order to set the aggregate public key in the frost-coordinator, not the address.

Closes: https://github.com/Trust-Machines/core-eng/issues/325


stjepangolemac (approved):
Can't vouch for Clarity correctness but other part looks okay. One suggestion I have is using a different value for public key. It doesn't look random and therefore might convey the wrong meaning. It's a real nitpick so feel free to ignore.

jferrant:
> Afaik you can get the public key from a Bech32m address, but sure. Let's just store the key directly, and leave it up to any caller to convert it into an address if they need to.

I was told this is not possible by Igor and by Chat GPT

netrome:
> > Afaik you can get the public key from a Bech32m address, but sure. Let's just store the key directly, and leave it up to any caller to convert it into an address if they need to.
> 
> I was told this is not possible by Igor and by Chat GPT

I feel outnumbered, but I'm pretty sure Igor and Chat GPT are wrong in this case. While it's true that generally speaking you can't typically get a public key from a bitcoin address as the address is commonly a hash of the public key or some script. However, in our case you can since we're using a taproot address. The taproot address is either a plain public key, or the root of a merkle tree denoting a script hash path. In our case, we're using the public key spend path which means the public key should be directly decodeable from the bech32m address.

netrome:
> > > Afaik you can get the public key from a Bech32m address, but sure. Let's just store the key directly, and leave it up to any caller to convert it into an address if they need to.
> > 
> > 
> > I was told this is not possible by Igor and by Chat GPT
> 
> I feel outnumbered, but I'm pretty sure Igor and Chat GPT are wrong in this case. While it's true that generally speaking you can't typically get a public key from a bitcoin address as the address is commonly a hash of the public key or some script. However, in our case you can since we're using a taproot address it **is possible.** The taproot address is either a plain public key, or the root of a merkle tree denoting a script hash path. In our case, we're using the public key spend path which means the public key should be directly decodeable from the bech32m address.

That being said, just storing and sending the public key directly is definitely easier for us - and I don't se a strong case for using the address here instead.

sbtc-ops/clarinet/tests/sbtc-alpha_test.ts
```diff
@@ -102,16 +102,16 @@ Clarinet.test({
         receipt.result.expectOk().expectBool(true);
 
         block = chain.mineBlock([
-            Tx.contractCall("sbtc-alpha", "set-bitcoin-wallet-address", [types.ascii("123456780abcdefghijklmnopqrstuvwxyz")], deployer.address),
+            Tx.contractCall("sbtc-alpha", "set-bitcoin-wallet-public-key", [types.buff(0x000000000000000000000000000000000000000000000000000000000000000000)], deployer.address),
```
igorsyl, line 105: Use a non-trivial test key?

jferrant: I don't think this is non trivial. It is a getter and a setter for an optional. If it exists. It was successful. However, not a big deal to change either.

xoloki (approved):
LGTM

PR #332, opened 2023-05-09
Add integration test for coordinator that runs `relay-server` `frost-signer` and `frost-coordinator`

xoloki:
Rebased on main with #331


xoloki:
Something is really broken with `codecov`, all I added was test code and it says test coverage dropped :P

.github/workflows/rust.yml
```diff
@@ -59,6 +59,8 @@ jobs:
         name: Cargo.lock
     - name: Cache JS script dependencies
       run: deno cache ./yarpc/js/stacks/transactions.ts
+    - name: Build code
+      run: cargo build --verbose --locked ${{matrix.type}}
```
netrome, line 63: Isn't this already covered by the `needs: build` directive?

frost-coordinator/Cargo.toml
```diff
@@ -20,6 +20,10 @@ tracing-subscriber = { workspace = true }
 frost-signer = { version = "0.0.1", path = "../frost-signer" }
 serde = { version = "1.0", features = ["serde_derive"] }
 
+[dev-dependencies]
+relay-server = { version = "0.1.0", path = "../relay-server" }
```
netrome, line 24: I suggest omitting the version for local dependencies in the same repo.
```suggestion
relay-server = { path = "../relay-server" }
```

netrome (approved):
Nice to see more integration tests! :rocket:

.github/workflows/rust.yml
xoloki: No, the build directive doesn't know the matrix, so it only builds debug

frost-coordinator/Cargo.toml
xoloki: Good catch!

xoloki: Done.

.github/workflows/rust.yml
netrome: Alright, thanks for clarifying!

PR #336, opened 2023-05-12
Encrypt DKG private shares

xoloki:
The last missing piece for `FROST` was to encrypt DKG private shares.  This PR uses AES-GCM to encrypt all outgoing DKG private shares, and decrypt incoming DKG private shares for the `key_ids` owned by each `Signer`.


stjepangolemac (approved):
Didn't spot anything wrong, at the surface crypto looks okay to me

frost-signer/src/signing_round.rs
```diff
@@ -387,21 +404,75 @@ impl SigningRound {
...
+        let mut invalid_dkg_private_shares = Vec::new();
+
+        for (src_key_id, encrypted_shares) in &self.shares {
+            let mut decrypted_key_shares = HashMap::new();
+
+            for (dst_key_id, private_share) in encrypted_shares {
+                if key_ids.contains(dst_key_id) {
+                    debug!(
+                        "decrypting dkg private share for key_id #{}",
+                        dst_key_id + 1
```
jferrant, line 420: wouldn't this just be decrypting dkg private share for dst_key_id? Is the +1 still needed?

frost-signer/src/signing_round.rs
```diff
@@ -557,13 +628,47 @@ impl SigningRound {
...
-            info!("sending dkg private share for key_id #{}", key_id);
+            info!(
+                "signer {} sending dkg private share for key_id #{}",
+                self.signer.signer_id, key_id
+            );
+            // encrypt each share for the recipient
+            let mut encrypted_shares = HashMap::new();
+
+            for (dst_key_id, private_share) in private_shares {
+                debug!(
```
jferrant, line 642: Similar to the other nit, I recommend pulling this internal block into an encrypt_private_share() function for unit testing/easier groking

frost-signer/src/signing_round.rs
```diff
@@ -387,21 +404,75 @@ impl SigningRound {
...
+                    debug!(
+                        "decrypting dkg private share for key_id #{}",
+                        dst_key_id + 1
+                    );
+                    let compressed =
+                        Compressed::from(self.signer_keys.key_ids[&(src_key_id + 1)].to_bytes());
+                    let src_public_key = Point::try_from(&compressed).unwrap();
+                    let shared_key = self.network_private_key * src_public_key;
+                    let nonce_vec = private_share[0..12].to_vec();
+                    let cipher_vec = private_share[12..].to_vec();
```
netrome, line 427: I don't understand where the magic 12 comes from, can you use a named constant for it? Also, is it guaranteed throughout the code that the private shares is always longer than 12 so we know this won't panic?

frost-signer/src/signing_round.rs
```diff
@@ -387,21 +404,75 @@ impl SigningRound {
...
+                        },
+                        Err(e) => {
+                            warn!("Failed to decrypt dkg private share from key_id {} to key_id {}: {:?}", src_key_id, dst_key_id, e);
+                            invalid_dkg_private_shares.push(*src_key_id);
+                        }
+                    }
+                } else {
+                    decrypted_key_shares.insert(*dst_key_id, Scalar::new());
+                }
+            }
```
netrome, line 448: +1 on @jferrant's comment to break this out.

frost-signer/src/signing_round.rs
```diff
@@ -557,13 +628,47 @@ impl SigningRound {
...
+                let compressed =
+                    Compressed::from(self.signer_keys.key_ids[&(dst_key_id + 1)].to_bytes());
+                let dst_public_key = Point::try_from(&compressed).unwrap();
+                let shared_key = self.network_private_key * dst_public_key;
+                let mut nonce_bytes = [0u8; 12];
+
+                rng.fill_bytes(&mut nonce_bytes);
+
+                let nonce_vec = nonce_bytes.to_vec();
+                let nonce = Nonce::from_slice(&nonce_vec);
```
netrome, line 655: The intermediary Vec seems unnecessary here:
```suggestion
                let nonce = Nonce::from_slice(&nonce_bytes);
```

netrome (approved):

.github/workflows/rust.yml
```diff
@@ -96,7 +96,7 @@ jobs:
         run: deno cache ./yarpc/js/stacks/transactions.ts
       - run: cargo install cargo-tarpaulin
       # - run: cargo tarpaulin
-      - run: cargo tarpaulin --verbose --all-features --workspace --timeout 120 --out Xml --avoid-cfg-tarpaulin
+      - run: cargo tarpaulin --verbose --all-features --workspace --timeout 120 --out Xml --avoid-cfg-tarpaulin --exclude frost-coordinator
```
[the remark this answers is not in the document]
jferrant: It breaks it at the moment. If you run cargo tarpualin localling it appears to clean the build dir. And build incrementally. I.e. it builds frost-signer then tests. Then frost-coordinator and tests. But it never builds the relay so it fails to find the binary within the test. (Whereas when you run cargo test, it builds the entire repo first and then executes each test). Not sure how to fix this :(

frost-signer/src/signing_round.rs
xoloki: Unfortunately, `+ 1` is still needed.  These `key_ids` are coming from `wsts` not `core-eng`, and we still haven't updated `wsts` to remove the `0` indexed `key_ids`.

frost-signer/src/signing_round.rs
```diff
@@ -387,21 +404,75 @@ impl SigningRound {
...
-            Ok(()) => DkgEnd {
-                dkg_id: self.dkg_id,
-                signer_id: self.signer.signer_id,
-                status: DkgStatus::Success,
-            },
-            Err(dkg_error_map) => DkgEnd {
+        let mut decrypted_shares = HashMap::new();
+
+        // go through private shares, and decrypt any for owned keys, leaving the rest as zero scalars
+        let key_ids: HashSet<u32> = self.signer.frost_signer.get_key_ids().into_iter().collect();
```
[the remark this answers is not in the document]
xoloki: Done, added `util::{make_shared_secret, encrypt, decrypt}` with tests.

frost-signer/src/signing_round.rs
xoloki: Done.

.github/workflows/rust.yml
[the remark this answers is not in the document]
xoloki: `frost-coordinator` only has the integration test, which as @jferrant noted breaks under `tarpaulin`.

frost-signer/src/signing_round.rs
xoloki: I had rather a large amount of trouble getting the `aes_gcm` code to work, which is why I made the nonce an explicit `Vec`.  I think it's helpful for didactic purposes.

.github/workflows/rust.yml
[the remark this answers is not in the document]
xoloki: > Should we add 'relay-server' as a build dependency to 'frost-coordinator'?

We do.  I tried it both as a regular dependency and a dev dependency.  Didn't seem to help either way.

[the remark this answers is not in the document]
sergey-shandar: Ok, another option is to run `cargo run --bin relay-server`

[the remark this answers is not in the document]
sergey-shandar: Ignore my previous comment. I think, we need to move the test from `unit` test to E2E test. https://github.com/orgs/Trust-Machines/discussions/322

E2E test could be just a shell script.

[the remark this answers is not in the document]
xoloki: Agreed, E2E testing is a better place for this, since it's just running some binaries.  Out of scope for this PR though.

PR #337, opened 2023-05-12
wip: Working on unit test. Found a bug blocked on upstream implementation

netrome:


commit-reveal-reference/src/lib.rs
```diff
@@ -267,6 +267,41 @@ fn internal_key() -> UntweakedPublicKey {
...
+    #[test]
+    fn commit_should_return_a_valid_bitcoin_p2tr_over_p2sh_address() {
+        let mut rng = helpers::seeded_rng();
+        let data: [u8; 64] = helpers::random_bytes(&mut rng);
+        let revealer_key = helpers::random_key(&mut rng);
+        let reclaim_key = helpers::random_key(&mut rng);
+
+        let commit_address = commit(&data, &revealer_key, &reclaim_key);
+
+        // TODO: This is failing. We are dependent on https://github.com/rust-bitcoin/rust-bitcoin/issues/1851 for the fix
```
stjepangolemac, line 279: Does it mean this is impossible to do without a helper function from their side or is very inconvenient?

commit-reveal-reference/src/lib.rs
```diff
@@ -267,6 +267,41 @@ fn internal_key() -> UntweakedPublicKey {
 mod tests {
     use super::*;
 
+    #[test]
+    fn commit_should_return_a_valid_bitcoin_p2tr_over_p2sh_address() {
+        let mut rng = helpers::seeded_rng();
+        let data: [u8; 64] = helpers::random_bytes(&mut rng);
```
stjepangolemac, line 273: It's a nitpick but this might be simple enough to inline. Something like this might be clearer:

```
let data: Vec<u8> = std::iter::from_fn(|| rng.gen()).take(64).collect();
```

If you want to stick with an array we can append `.try_into().unwrap()` at the end of line above.

stjepangolemac (approved):

commit-reveal-reference/src/lib.rs
netrome: We can do it without their helper function. It would be more convenient to have it upstream though

commit-reveal-reference/src/lib.rs
netrome: I like using the array when we have explicit sizes. Here I think `[rng.gen(); N]` is the simplest. I'm okay with inlining it.

netrome: Updated.

PR #334, opened 2023-05-10
feat(test-vectors): generate test-vectors

soju-drinker:
Refactors how test-vectors are generated and adds new test-vectors for peg-in, peg-out request and peg-handoff.

Closes #302

PR #339, opened 2023-05-12
Introducing a library for creating commit addresses and reveal transaction

netrome:


commit-reveal-reference/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+    #[error("Could not build taproot spend info: {0}: {1}")]
+    InvalidTaproot(&'static str, TaprootBuilderError),
+}
+
+type CommitRevealResult<T> = Result<T, CommitRevealError>;
+
+fn peg_in_data(address: StacksAddress, contract_name: Option<String>, reveal_fee: u64) -> Vec<u8> {
+    let mut data: Vec<u8> = Vec::with_capacity(86);
+
+    data.push('<' as u8);
```
sergey-shandar, line 39: IMHO, we should use the [chain](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.chain) function to create an iterator and then use [collect](https://doc.rust-lang.org/nightly/std/iter/trait.Iterator.html#method.collect).

commit-reveal-reference/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+    data.extend_from_slice(address.bytes.as_bytes());
```
MarvinJanssen, line 40: Just checking, but `address.bytes.as_bytes()` will include the version byte, right?

MarvinJanssen (commented):
How does this PR relate to the test vectors in #334? I started using that PR to create Mini test transactions.

commit-reveal-reference/src/lib.rs
netrome: Agreed. Iterators are nicer to read. I'll update it.

commit-reveal-reference/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+
+    let contract_name_bytes = contract_name
+        .map(|contract_name| contract_name.as_bytes().to_vec())
```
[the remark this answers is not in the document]
netrome: Agreed. I have a ticket to introduce it. I want to add it to Stacks first to see if there's any pushback from the folks at their end.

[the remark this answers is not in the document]
netrome: Here's the issue: https://github.com/Trust-Machines/core-eng/issues/342

netrome:
> How does this PR relate to the test vectors in #334? I started using that PR to create Mini test transactions.

The test vectors is intended to produce serialized bitcoin transactions with the appropriate format, useful for e.g. Mini testing and unit tests. This PR intends to create a working commit-reveal flow, from generating a commit address to producing a valid reveal transaction based on this address. There's definitely some overlap in these two PRs, since they have been implemented in parallel, and it would be nice to factor out some common code when we have the time.

commit-reveal-reference/src/lib.rs
netrome: Done :heavy_check_mark:

commit-reveal-reference/src/lib.rs
netrome: Added :heavy_check_mark:

commit-reveal-reference/src/lib.rs
```diff
@@ -0,0 +1,354 @@
+// TODO: Before PR
+// - [x] Error handling
+// - Unit testing
+// - Rustdoc for functions and library
```
sergey-shandar, line 4: Add links to SIPs.

commit-reveal-reference/src/lib.rs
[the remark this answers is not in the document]
MarvinJanssen: I realised we are getting pretty close to the Stacks wire format for principals. If the payload were to just be wire format, then converting the buffer to a principal (either standard or contract) is just two lines of code: a buffer slice and a `from-consensus-buff?`. Definitely more runtime efficient although it will introduce another byte.

Taken from [SIP005](https://github.com/stacksgov/sips/blob/main/sips/sip-005/sip-005-blocks-and-transactions.md):

>The following type IDs indicate the following values:
>- 0x05: standard principal
>- 0x06: contract principal

>Standard principal
>
>The next byte is the address version, and the following 20 bytes are the principal's public key(s)' Hash160.
>
>Contract Principal
>
>The next byte is the address version, the following 20 bytes are a Hash160, the 21st byte is the length of the contract name, and the remaining bytes (up to 128, exclusive) encode the name itself.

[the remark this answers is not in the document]
netrome: Hmm... right. So that would be the type ID we're also introducing then? We do have the extra byte available, so if it's more runtime efficient I think we could use the wire format directly. It's definitely easier to explain.

A note here: We're capping the contract names at 40 bytes. I think that's the current max length, although 128 is supported in some places for legacy reasons.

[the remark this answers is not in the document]
netrome: I started a discussion in the current PR introducing length-prefixing in Stacks: https://github.com/stacks-network/stacks-blockchain/pull/3718

radicleart:
Is the goal to have these features behind stacks node api calls ? That would be very cool and take a load of complexity out of the clients. 

I have a WIP on a parallel flow in javascript which stores the taproot data in mongo db. Let me know if this can be of use - e.g. you can generate a commit tx in the bridge web app you can see all the taproot data via  a Bridge API call.

There are some caveats e.g. this [issue #130](https://github.com/Trust-Machines/sbtc-bridge-web/issues/130) about mixing secp and schnorr pub keys in taproot scripts (problem only for the reclaim path) - curious to know if you find similar issues with the rust libraries?

netrome:
> Is the goal to have these features behind stacks node api calls ? That would be very cool and take a load of complexity out of the clients.
> 
> I have a WIP on a parallel flow in javascript which stores the taproot data in mongo db. Let me know if this can be of use - e.g. you can generate a commit tx in the bridge web app you can see all the taproot data via a Bridge API call.
> 
> There are some caveats e.g. this [issue #130](https://github.com/Trust-Machines/sbtc-bridge-web/issues/130) about mixing secp and schnorr pub keys in taproot scripts (problem only for the reclaim path) - curious to know if you find similar issues with the rust libraries?

I was thinking more about using this as a library. However we'll need to expose some sort of API for the coordinator to post the reveal data from the committer. I know we've duplicated some commit logic here, mostly to help us ensure we get the reveal logic right since that's what we're going to deploy in the coordinator later.

It would make sense for us to sync the state of this and some challenges we're facing. I'll be off with the TrustMachines folks in Cancun next week but I'd love to catch up more afterwards.

sergey-shandar:
@netrome do you still need the branch?

PR #319, opened 2023-05-03
First version of sBTC Mini peg-in and registry

MarvinJanssen:
This branch will be used to implement the peg-in and any changes to the registry that are necessary.

Also adds `@mine-blocks-before` unit test annotation.

Testing for Contracts:
- sbtc-registry.clar
- sbtc-peg-in-processor.clar

PR #377, opened 2023-05-25
sBTC Mini Stacking Pool & Peg-Transfer 1st Drafts

setzeus:
1st drafts of both Clarity contracts named above. There are still significant unknowns & troublesome testing so there is no meaningful coverage at the moment - the point of this PR is more so a project draft completion & catch-up with the rest of sBTC-mini contracts.

**Stacking Pool**
The stacking pool implements the five-window mechanism laid out in the sBTC Master doc (disburse/register/vote/transfer/penalty) with every core function implemented (though again, without unit tests). 

**Peg Transfer / Handoff**
The peg-transfer / handoff contract contains a single function that verifies the peg-balance has been transferred from the previous threshold-wallet to the current threshold-wallet.


sbtc-mini/contracts/sbtc-peg-transfer.clar
```diff
@@ -0,0 +1,113 @@
...
+
+            ;; Done manually for read/write concerns
+            (tx-output-0 (default-to {value: u0, scriptPubKey: current-threshold-hashbytes} (element-at tx-outputs u0)))
+            (tx-output-1 (default-to {value: u0, scriptPubKey: current-threshold-hashbytes} (element-at tx-outputs u1)))
+            (tx-output-2 (default-to {value: u0, scriptPubKey: current-threshold-hashbytes} (element-at tx-outputs u2)))
+            (tx-output-3 (default-to {value: u0, scriptPubKey: current-threshold-hashbytes} (element-at tx-outputs u3)))
+            (tx-output-4 (default-to {value: u0, scriptPubKey: current-threshold-hashbytes} (element-at tx-outputs u4)))
+            (tx-output-5 (default-to {value: u0, scriptPubKey: current-threshold-hashbytes} (element-at tx-outputs u5)))
+            (tx-output-6 (default-to {value: u0, scriptPubKey: current-threshold-hashbytes} (element-at tx-outputs u6)))
+            (tx-output-7 (default-to {value: u0, scriptPubKey: current-threshold-hashbytes} (element-at tx-outputs u7)))
```
MarvinJanssen, line 71: We can probably optimise this in a future PR.

sbtc-mini/contracts/sbtc-peg-transfer.clar
```diff
@@ -0,0 +1,113 @@
...
+        
+            ;; Assert we're in the transfer window
+            (asserts! (is-eq (contract-call? .sbtc-stacking-pool get-current-window) "transfer")  err-not-in-transfer-window)
+
+            ;; Assert that balance of previous-threshold-wallet wasn't already transferred
+            (asserts! previous-pool-balance-transferred err-balance-already-transferred)
+
+            ;; Assert that every unwrapped receiver addresss is equal to new/current-threshold-wallet
+            (asserts! (and 
+                (is-eq current-threshold-hashbytes (get scriptPubKey tx-output-0))
```
MarvinJanssen, line 85: This is not going to work, the hash bytes + version need to be formatted as a scriptPubKey.

sbtc-mini/contracts/sbtc-peg-transfer.clar
```diff
@@ -0,0 +1,113 @@
...
+                (is-eq current-threshold-hashbytes (get scriptPubKey tx-output-3))
+                (is-eq current-threshold-hashbytes (get scriptPubKey tx-output-4))
+                (is-eq current-threshold-hashbytes (get scriptPubKey tx-output-5))
+                (is-eq current-threshold-hashbytes (get scriptPubKey tx-output-6))
+                (is-eq current-threshold-hashbytes (get scriptPubKey tx-output-7))
+            ) err-wrong-pubkey)
+
+            ;; Assert that amount transferred > sbtc-registry peg-balance
+            (asserts! 
+                (< cycle-peg-balance
```
MarvinJanssen, line 97: Should be `<=`.

sbtc-mini/contracts/sbtc-peg-transfer.clar
```diff
@@ -0,0 +1,113 @@
...
+                        (get value tx-output-0)
+                        (get value tx-output-1)
+                        (get value tx-output-2)
+                        (get value tx-output-3)
+                        (get value tx-output-4)
+                        (get value tx-output-5)
+                        (get value tx-output-6)
+                        (get value tx-output-7)
+                    )
+                ) err-peg-balance-not-sufficient)
```
MarvinJanssen, line 108: I think this design is a bit brittle. If the signers, for whatever reason, do the hand-off in more than one transaction, then the protocol is stuck. (Forever?)

sbtc-mini/contracts/sbtc-registry.clar
```diff
@@ -117,3 +150,12 @@
...
 )
+
+;; Called through the sbtc-stacking-pool contract
+(define-public (penalty-peg-state-change)
+	(begin 
+		(try! (is-protocol-caller contract-caller))
+		;; not sure if additional checks needed
+		(ok (var-set peg-state false))
+	)
+)
```
MarvinJanssen, line 161: Why is this function necessary when `set-peg-state-exists`?

Also, do you think a `peg-state` variable is necessary? Can it not be derived from the registry state?

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -1,40 +1,23 @@
...
-;; Same burnchain and PoX constants as mainnet
-(define-data-var first-burn-block-height uint u666050)
-(define-data-var reward-cycle-len uint u2100)
-;; Relative burnchain block heights (between 0 and 2100) as to when the system transitions into different states
-(define-data-var registration-window-rel-end uint u1600)
-(define-data-var voting-window-rel-end uint u1900)
-(define-data-var transfer-window-rel-end uint u2000)
-(define-data-var penalty-window-rel-end uint u2100)
+;; Burn POX address for penalizing stackers/signers
+(define-constant pox-burn-address { version: 0x00, hashbytes: 0x0011223344556699001122334455669900112233445566990011223344556699})
```
MarvinJanssen, line 15: Probably want to just use the zero address.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -55,27 +38,55 @@
...
 ;; Current cycle threshold wallet
 (define-data-var threshold-wallet { version: (buff 1), hashbytes: (buff 32) } { version: 0x00, hashbytes: 0x00 })
 
-;;;;;;;;;;;;
+;; Current signer minimal
+(define-data-var signer-minimal uint u0)
+
+;; Same burnchain and PoX constants as mainnet
+(define-data-var first-burn-block-height uint u666050)
+(define-data-var reward-cycle-len uint u2100)
```
MarvinJanssen, line 71: These can be constants.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -117,10 +128,7 @@
 (define-map payout-address-in-cycle { version: (buff 1), hashbytes: (buff 32) } uint)
 
 
-
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;;;; Read-Only Functions ;;;;;
-;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 ;; Get current cycle pool
 (define-read-only (get-current-cycle-pool)
```
MarvinJanssen, line 134: You'll have to upgrade to pox-3 now. 😁

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -148,32 +178,117 @@
...
-                    "registration"
-                    (if (and (>= burn-block-height start-voting-window) (< burn-block-height start-transfer-window))
-                        "voting"
-                        (if (and (>= burn-block-height start-transfer-window) (< burn-block-height start-penalty-window))
-                            "transfer"
-                            "penalty"
-                        )
+        (asserts! peg-state "bad-peg")
+
+        (if (< latest-disbursed-burn-height burn-block-height)
```
MarvinJanssen, line 183: Can refactor the nested ifs to a series of `asserts!`. Also, using constants with 1 byte buffer values will be more runtime efficient than using `string-ascii`.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -148,32 +178,117 @@
...
+
+            ;; Assert that rewards haven't already been disbursed
+            (asserts! (not previous-pool-disbursed) err-already-disbursed)
+
+            ;; Assert that transaction was mined...tbd last two params
+            (unwrap! (contract-call? .clarity-bitcoin was-segwit-tx-mined-compact burn-height tx header tx-index tree-depth wproof 0x 0x ctx cproof) err-tx-not-mined)
+
+            ;; Assert that every unwrapped receiver addresss is equal to previous-threshold-wallet
+            (asserts! (and 
+                (is-eq previous-threshold-wallet (get scriptPubKey tx-output-0))
```
MarvinJanssen, line 257: Same comment as the hand-off contract, the hash bytes + version needs to be formatted as a scriptPubKey.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -148,32 +178,117 @@
...
+	(tx-index uint)
+	(tree-depth uint)
+	(wproof (list 14 (buff 32)))
+    (witness-merkle-root (buff 32))
+    (witness-reserved-data (buff 32))
+	(ctx (buff 1024))
+	(cproof (list 14 (buff 32))))
+    (let 
+        (
+            (current-cycle (contract-call? 'SP000000000000000000002Q6VF78.pox-2 current-pox-reward-cycle))
```
MarvinJanssen, line 223: If the proving takes too long then pox will move to the next reward cycle and the function will fail. I think the function should accept a cycle parameter.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -343,6 +454,7 @@
                     (unwrapped-candidate (unwrap! current-candidate-status err-unwrapping-candidate))
```
MarvinJanssen, line 454: Can this large if block be simplified with `default-to`?

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -357,9 +469,22 @@
                 ;; Update signer map
                 (map-set signer {stacker: tx-sender, pool: next-cycle} (merge next-pool-signer { vote: (some pox-addr) }))
 
-                ;; Update pool map
-
-                ;; still a few to-dos here...
+                ;; Check if 70% wallet consensus has been reached
+                (if (>= (/ (* new-candidate-votes u100) next-pool-total-stacked) u70)
```
MarvinJanssen, line 473: Perhaps it would be nice to be able to change the threshold in the future? (Via a `is-protocol-caller` guarded setter.) If so, then it would also be prudent to add a bit more precision. cc @jo-tm

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -370,7 +495,175 @@
...
+            (next-cycle (+ current-cycle u1))
+            (current-pool (unwrap! (map-get? pool current-cycle) err-pool-cycle))
+            (current-pool-stackers (get stackers current-pool))
+            (next-pool (unwrap! (map-get? pool next-cycle) err-pool-cycle))
+        )
+
+        ;; Assert that we're in the transfer window
+        (asserts! (is-eq (get-current-window) "transfer") err-not-in-transfer-window)
+
+        ;; Assert that pending-wallet-peg-outs is equal to zero
```
MarvinJanssen, line 538: Did you mean "is not equal to zero"?

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -370,7 +495,175 @@
...
+        (asserts! (> (contract-call? .sbtc-registry get-pending-wallet-peg-outs) u1) err-unhandled-request)
```
MarvinJanssen, line 539: Did you mean `> u0`?

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -370,7 +495,175 @@
...
+                (map-set pool next-cycle (merge
+                    next-pool
+                    {last-aggregation: (some block-height), reward-index: (some ok-result)}
+                ))
+            err-result
+                false
+        )
+
+        ;; Change peg-state to "bad-peg"
+        (ok (unwrap! (contract-call? .sbtc-registry penalty-peg-state-change) (err u1)))
```
MarvinJanssen, line 591: Can just end with the contract call. (Instead of `ok` and `unwrap!`.)

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -370,7 +495,175 @@
...
+        ;; Penalize stackers by re-stacking but with a pox-reward address of burn address
+        (match (as-contract (contract-call? 'SP000000000000000000002Q6VF78.pox-2 stack-aggregation-commit-indexed pox-burn-address next-cycle))
+            ok-result
+                (map-set pool next-cycle (merge
+                    next-pool
+                    {last-aggregation: (some block-height), reward-index: (some ok-result)}
+                ))
+            err-result
+                false
+        )
```
MarvinJanssen, line 588: Repetition can be refactored into a private function. (cf. `penalty-unhandled-request`.)

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -370,7 +495,175 @@
...
+
+        ;; Change peg-state to "bad-peg"
+        (ok (unwrap! (contract-call? .sbtc-registry penalty-peg-state-change) (err u1)))
```
MarvinJanssen, line 630: As above.

MarvinJanssen (commented):
Nice, lots of progress.

sbtc-mini/contracts/sbtc-peg-transfer.clar
setzeus: Copy. Believe I went with this after seeing that mapping/folding was more inefficient than doing it manually.

MarvinJanssen: A fold will definitely be more expensive, just wondering if you want to map all outputs this way or sum them. I think the setup is pretty efficient if this is what you want to do.

sbtc-mini/contracts/sbtc-peg-transfer.clar
setzeus: Can you clarify? Thought the version was an internal Clarity struct to help with formatting by defining the transaction type & hashbytes was the actual key?

sbtc-mini/contracts/sbtc-peg-transfer.clar
setzeus: Thx, fixed.

sbtc-mini/contracts/sbtc-peg-transfer.clar
setzeus: In agreement. Most definitely don't think this is the final design but wanted to get a draft up there / caught-up with the rest of the BTC processors.

sbtc-mini/contracts/sbtc-peg-transfer.clar
FriendsFerdinand: No, the version is the address type. See how to turn hashbytes into ScriptPubKey (@MarvinJanssen if you can double-check pls):
https://gist.github.com/FriendsFerdinand/5a74aa1f3a3c27408efcbd7419e4007e

sbtc-mini/contracts/sbtc-peg-transfer.clar
```diff
@@ -0,0 +1,113 @@
...
+	(tree-depth uint)
+	(wproof (list 14 (buff 32)))
+    (witness-merkle-root (buff 32))
+    (witness-reserved-data (buff 32))
+	(ctx (buff 1024))
+	(cproof (list 14 (buff 32))))
+    (let 
+        (
+            (current-cycle (contract-call? 'SP000000000000000000002Q6VF78.pox-2 current-pox-reward-cycle))
+            (current-pool-unwrapped (unwrap! (contract-call? .sbtc-stacking-pool get-current-cycle-pool) err-current-pool-not-found))
```
FriendsFerdinand, line 50: I think we want to use `get-specific-cycle-pool` here because we already have the current-cycle and `get-current-cycle-pool` does the same thing too.

sbtc-mini/contracts/sbtc-peg-transfer.clar
```diff
@@ -0,0 +1,113 @@
...
+            (current-threshold-wallet (unwrap! (get threshold-wallet current-pool-unwrapped) err-current-threshold-wallet))
+            (current-threshold-hashbytes (get hashbytes current-threshold-wallet))
+
+            (previous-cycle (- current-cycle u1))
+            (previous-pool-unwrapped (unwrap! (contract-call? .sbtc-stacking-pool get-specific-cycle-pool previous-cycle) err-previous-pool-not-found))
+            (previous-threshold-wallet (get hashbytes (unwrap! (get threshold-wallet previous-pool-unwrapped) err-previous-threshold-wallet)))
+            (previous-pool-balance-transferred (get balance-transferred previous-pool-unwrapped))
+
+            (cycle-peg-balance (contract-call? .sbtc-registry get-peg-balance))
+            (parsed-tx (unwrap! (contract-call? .clarity-bitcoin parse-tx tx) err-parsing-btc-tx))
```
FriendsFerdinand, line 60: Presumably, this transaction is a segwit transaction? In this case, `parse-wtx` has to be used. If the transaction can be either type, then we'll have to use an `if` branch

sbtc-mini/contracts/sbtc-peg-transfer.clar
FriendsFerdinand: think, we can just do a lazy eval

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -132,12 +140,34 @@
...
 ;; Get current window
 (define-read-only (get-current-window)
     (let 
         (
             ;; to-do -> (get-peg-state) from .sbtc-controller, returns bool
-            (peg-state true)
+            (peg-state (contract-call? .sbtc-registry current-peg-state))
             (current-cycle (contract-call? 'SP000000000000000000002Q6VF78.pox-2 current-pox-reward-cycle))
             (current-cycle-burn-height (contract-call? 'SP000000000000000000002Q6VF78.pox-2 reward-cycle-to-burn-height current-cycle))
             (next-cycle (contract-call? 'SP000000000000000000002Q6VF78.pox-2 current-pox-reward-cycle))
```
FriendsFerdinand, line 173: `next-cycle` and `current-cycle` are the same here.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -148,32 +178,117 @@
...
+                (is-eq previous-threshold-wallet (get scriptPubKey tx-output-3))
+                (is-eq previous-threshold-wallet (get scriptPubKey tx-output-4))
+                (is-eq previous-threshold-wallet (get scriptPubKey tx-output-5))
+                (is-eq previous-threshold-wallet (get scriptPubKey tx-output-6))
+                (is-eq previous-threshold-wallet (get scriptPubKey tx-output-7))
+            ) err-wrong-pubkey)
+
+            ;; Assert that every unwrapped output is less than dust
+            (asserts! (and 
+                (< (get value tx-output-0) dust-limit)
```
FriendsFerdinand, line 269: I'm unsure as to how this checks that the funds that went to the ts wallet, went to the previous pool signers. Are you assuming that these outputs are like Accounts that are increased and decreased as funds are spent by the address?

sbtc-mini/contracts/sbtc-registry.clar
setzeus: Updated.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Updated, ty.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: The first one yeap. 

The second I'm not as sure? I suppose that's an open question: do we want pool cycles & POX cycles to _always_ stay in-sync? What does this mean when thrown into a bad-state?

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -224,16 +341,12 @@
...
-
         ;; Record pre-signer
         (ok (map-set pre-signer {stacker: tx-sender, pool: next-cycle} true))
 
     )
 )
 
-;; Register as a signer
 ;; @desc: registers a signer for the cycle, goal of this function is to gurantee the amount of STX to be stacked for the next cycle
 (define-public (signer-register (pre-registered-signer principal) (amount-ustx uint) (pox-addr { version: (buff 1), hashbytes: (buff 32)}) (public-key (buff 33)))
```
FriendsFerdinand, line 351: `(public-key (buff 33))`

aren't public keys for schnorr signatures 32 bytes?

FriendsFerdinand:
Here:
https://github.com/Trust-Machines/core-eng/blob/411c05d1042205a3c89098fb8dc903fe9f51199a/sbtc-mini/contracts/sbtc-stacking-pool.clar#L447
You're using `as-max-len?` with 10, but the maximum amount in the map is 100. I'm assuming, this is a mistake

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -299,12 +409,13 @@
             (current-candidate-status (map-get? votes-per-cycle {cycle: next-cycle, wallet-candidate: pox-addr}))
             (next-pool (unwrap! (map-get? pool next-cycle) err-pool-cycle))
             (next-threshold-wallet (get threshold-wallet next-pool))
+            (next-pool-total-stacked (get stacked next-pool))
             (next-pool-signer (unwrap! (map-get? signer {stacker: tx-sender, pool: next-cycle}) err-not-signer))
```
FriendsFerdinand, line 413: Will this ever pass? I find the signer is only defined (`map-set signer`) in this function after this line and not anywhere else.

sbtc-mini/contracts/sbtc-peg-transfer.clar
MarvinJanssen: @FriendsFerdinand looks correct. I also have a function to turn the version + hash bytes into a scriptPubKey in my PR #319. (Direct link: https://github.com/Trust-Machines/core-eng/blob/2434a7b3031813e3e01b64147c02366c4fcca77f/sbtc-mini/contracts/sbtc-btc-tx-helper.clar#L75-L80.)

Yes, always P2TR but I also support P2WPKH and P2WSH for now because it made testing easier. Simple to slash that in the future. I purposely did not add the logic to the registry to make it more agnostic. If a future format is added then we can upgrade the helper.

sbtc-mini/contracts/sbtc-peg-transfer.clar
MarvinJanssen: Yep it's what it do to extract/process data as well.

sbtc-mini/contracts/sbtc-stacking-pool.clar
MarvinJanssen: I guess you can call `get-pox-info` on the PoX contract to get the cycle length and then store it in a data var. It is pretty much static so calling PoX every time would be a waste of runtime. 

I can't answer your questions though, something to bring up during the next meeting. I assumed sBTC cycles were always in sync with PoX.

sbtc-mini/contracts/sbtc-peg-transfer.clar
MarvinJanssen: Under normal circumstances it will be a segwit transaction. But I think according to the spec, the hand-off does not _have_ to happen by the signers. Theoretically, anyone could fund the new address with the right amount of Bitcoin and the protocol would recover. (The same applies to peg out, anyone can fulfil them. So if something terrible happens, the signers or someone else can decide to fulfil a request outside of the protocol.)

sbtc-mini/contracts/sbtc-stacking-pool.clar
MarvinJanssen: That's right.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: I don't believe so since there are two functions that are specific to that branch? Votes-per-cycle & next-cycle both have unique updates based on whether the candidate exists? 

Could refactor to a "match" but unclear on how default-to would fulfill both map-sets.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Great idea. Did the following:

Updated to 3 digits / divide by 1000, created a threshold-consensus variable, & created an update-threshold-that checks with a read-only (is-protocol-caller ...) similar to yours.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Copy. Refactored four functions into a private helper

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Great catch. You're right, should be in (signer-register ...) as well.

sbtc-mini/contracts/sbtc-peg-transfer.clar
jo-tm: lets open a issue for this question, if this is not in the MVP critical path.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Check inputs were spent or outputs were sent to POX addresses.***

sbtc-mini/.DS_Store
MarvinJanssen, line 1: You accidentally committed a `.DS_Store` file.

Strange I thought it was listed in `.gitignore`.

sbtc-mini/contracts/pox-2.clar
MarvinJanssen, line 1: Can we remove PoX-2?

sbtc-mini/tests/sbtc-stacking-pool_test.clar
```diff
@@ -13,24 +13,21 @@
...
 
 ;; @name Get current window
 ;; @mine-blocks-before 2100
 (define-public (test-get-current-window)
-    (begin
-		(asserts!
-			(is-eq "penalty" (contract-call? .sbtc-stacking-pool get-current-window))
-			(err "Should have succeeded")
-		)
+	(if (is-eq 0x00 (contract-call? .sbtc-stacking-pool get-current-window))
```
MarvinJanssen, line 29: Safer to use constant instead of `0x00` magic value.

sbtc-mini/contracts/sbtc-peg-transfer.clar
```diff
@@ -69,32 +71,35 @@
...
+            (current-unwrapped-threshold-pubkey (unwrap! (contract-call? .sbtc-btc-tx-helper hashbytes-to-scriptpubkey current-threshold-wallet) err-threshold-to-scriptpubkey))
         )
 
             ;; Assert that transaction was mined...tbd last two params
-            (unwrap! (contract-call? .clarity-bitcoin was-segwit-tx-mined-compact burn-height tx header tx-index tree-depth wproof 0x 0x ctx cproof) err-tx-not-mined)
+            (unwrap! (contract-call? .clarity-bitcoin was-segwit-tx-mined-compact burn-height tx header tx-index tree-depth wproof witness-merkle-root witness-reserved-data ctx cproof) err-tx-not-mined)
         
             ;; Assert we're in the transfer window
-            (asserts! (is-eq (contract-call? .sbtc-stacking-pool get-current-window) "transfer")  err-not-in-transfer-window)
+            (asserts! (is-eq (contract-call? .sbtc-stacking-pool get-current-window) 0x03)  err-not-in-transfer-window)
```
MarvinJanssen, line 83: Best to use constant instead of magic value `0x03`.

sbtc-mini/contracts/sbtc-peg-transfer.clar
```diff
@@ -69,32 +71,35 @@
             (tx-output-5 (default-to {value: u0, scriptPubKey: current-threshold-hashbytes} (element-at tx-outputs u5)))
             (tx-output-6 (default-to {value: u0, scriptPubKey: current-threshold-hashbytes} (element-at tx-outputs u6)))
             (tx-output-7 (default-to {value: u0, scriptPubKey: current-threshold-hashbytes} (element-at tx-outputs u7)))
+
+            ;; versions + hashbytes to scriptPubKey
+            (current-unwrapped-threshold-pubkey (unwrap! (contract-call? .sbtc-btc-tx-helper hashbytes-to-scriptpubkey current-threshold-wallet) err-threshold-to-scriptpubkey))
```
MarvinJanssen, line 76: What happens if the peg transfer happens too late and we enter the next cycle? I think it is a question we have been having at different parts of the protocol. Let's discuss in the next meeting and create an issue.

sbtc-mini/tests/pox-3_test.ts
MarvinJanssen, line 1: File can be deleted.

sbtc-mini/contracts/sbtc-peg-transfer.clar
setzeus: Added same window constants used in stacking-pool

sbtc-mini/contracts/sbtc-peg-transfer.clar
setzeus: Copy. Will draft issue & surface on our T meeting.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Just checked & realized I overlooked this in latest commit so leaving unresolved momentarily.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Realizing this is a difference b/w mini & production, closing for now but will create issue about this.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Fixed

setzeus:
Querying for approvals, @MarvinJanssen & @FriendsFerdinand - lmk if you guys have any additional comments / changes needed.

@jo-tm & @radicleart feel free to review yourselves too if / when you have the time.

FriendsFerdinand (approved):

sbtc-mini/contracts/sbtc-btc-tx-helper.clar
```diff
@@ -0,0 +1,86 @@
...
+	(wproof (list 14 (buff 32))) ;; merkle proof for wtxids
+	(witness-merkle-root (buff 32))
+	(witness-reserved-data (buff 32))
+	(ctx (buff 1024)) ;; non-segwit coinbase tx, contains the witness root hash
+	(cproof (list 14 (buff 32))) ;; merkle proof for coinbase tx
+	;; proof and cproof trees could somehow be condensed into a single list
+	;; because they converge at some point
+	)
+	(begin
+		;; TODO: optimise to one call
```
jo-tm, line 23: let's remove TODOs from main branch and open a new issue.

jo-tm (changes requested):
small TODO removal and one test is failing (if is not critical just by pass it): `, should return none ... ok (57ms)`

`/.test/sbtc-stacking-pool_test.ts => test-get-current-window: Get current window ... FAILED (101ms)`

jo-tm (approved):

PR #379, opened 2023-05-25
Stacks coordinator - parsing responses from stacks-node

stjepangolemac:
This successfully gets the error description even though it's a bit crude. At this moment I'm not sure why parsing it as JSON fails. I've checked and response contains the JSON content type header.

@jferrant if you have time feel free to assign yourself too and see what is causing this.


stjepangolemac:
We were running main on the vm. Will close this.

PR #386, opened 2023-05-30
Change coordinator integration test to use Rust objects not binaries

xoloki:
The current coordinator integration test runs `relay-server`, `frost-coordinator`, and `frost-signer` binaries.  This causes problems with some `codecov` tooling, since the binaries are not present in that build.

This PR changes the integration test to instead create Rust objects for the various components.  This also allows us to dynamically generate configs using fresh keys, for any number of signers.


xoloki:
Ok, this is truly bizarre.  We're getting failed signature verifications on DkgPrivateShares, but only in certain cases.

If I play with `num_signers`, everything from `1` to `12` works *except* `7`.  And everything `> 12` fails.

I first thought something was getting messed up with the keys, so I changed it to use the same key in coordinator and all signers.  Same error.

I then thought maybe there was something weird going on with threads, so I tried forking instead of using threads.  Same error!

Gonna keep digging....

xoloki:
Super weird.  I made a new `signer7.toml` with 7 signers and coordinator all having the same keys, then ran it under `screen`.  Same error!

xoloki:
Coverage has now increased from `49.52%` to `65.35%`... let there be much rejoicing.

frost-coordinator/src/coordinator.rs
```diff
@@ -469,51 +469,131 @@ where
 
 #[cfg(test)]
 mod test {
+    use crate::{
+        coordinator::{Command, Coordinator},
+        DEVNET_COORDINATOR_ID,
+    };
```
netrome, line 475: Nit: When test modules are in the same file I prefer to import super::*. It saves a lot of import lines, and since everything is declared in the same file it does not hurt readability.
```suggestion
    use super::*;
```

frost-coordinator/src/coordinator.rs
```diff
@@ -469,51 +469,131 @@ where
...
+    use rand_core::OsRng;
+    use relay_server::Server as RelayServer;
+    use std::{env, thread};
+
+    fn create_signer_key_ids(signer_id: u32, keys_per_signer: u32) -> Vec<u32> {
+        let mut signer_key_ids = Vec::new();
+        for i in 0..keys_per_signer {
+            signer_key_ids.push(keys_per_signer * signer_id + i + 1);
+        }
+        signer_key_ids
```
netrome, line 493: Nit: imo it's cleaner to use iterators whenever possible to construct collections
```suggestion
        (0..keys_per_signer)
            .map(|i| keys_per_signer * signer_id + i + 1)
            .collect()
```

frost-coordinator/src/coordinator.rs
```diff
@@ -469,51 +469,131 @@ where
...
+                )
+            })
+            .collect::<HashMap<u32, ecdsa::PublicKey>>();
+
+        let mut key_ids = HashMap::new();
+        for (signer_id, signer_key) in &signer_id_keys {
+            for i in 0..keys_per_signer {
+                key_ids.insert(3 * signer_id - i, signer_key.clone());
+            }
+        }
```
netrome, line 511: Nit: Same here, this is slightly cleaner (also avoiding vector indexing by using `.enumerate`)
```suggestion
        let signer_id_keys: HashMap<u32, ecdsa::PublicKey> = signer_private_keys
            .iter()
            .enumerate()
            .map(|(i, key)| ((i + 1) as u32, ecdsa::PublicKey::new(key).unwrap()))
            .collect();

        let key_ids: HashMap<u32, ecdsa::PublicKey> = signer_id_keys
            .iter()
            .flat_map(|(signer_id, signer_key)| {
                (0..keys_per_signer).map(|i| (3 * *signer_id - i, signer_key.clone()))
            })
            .collect();
```

frost-coordinator/src/coordinator.rs
```diff
@@ -469,51 +469,131 @@ where
...
-            &log_info,
+        let keys_threshold: u32 = match env::var("keys_threshold") {
+            Ok(keys_threshold) => keys_threshold.parse::<u32>().unwrap(),
+            Err(_) => 10,
+        };
+        let keys_per_signer: u32 = match env::var("keys_per_signer") {
+            Ok(keys_per_signer) => keys_per_signer.parse::<u32>().unwrap(),
+            Err(_) => 3,
+        };
+        let mut rng = OsRng::default();
```
netrome, line 538: I would prefer to use a seedable rng here, such as `rand::StdRng` https://rust-random.github.io/rand/rand/rngs/struct.StdRng.html. Optionally, the random seed could be sampled using OsRng as long as it is logged so any flaky behavior can be reproduced.

netrome (approved):
Good stuff. Some smaller comments from me.

frost-coordinator/src/coordinator.rs
xoloki: Done.  I had to import DEVNET_COORDINATOR_ID separately.

frost-coordinator/src/coordinator.rs
xoloki: Done.

frost-coordinator/src/coordinator.rs
xoloki: Done.

frost-coordinator/src/coordinator.rs
xoloki: Done.

xoloki:
The bug has been squashed; it turns out that when HashMap says its iteration is unordered, it really means unordered, even comparing two HashMaps of the same size on the same machine.  So when rcpt hashed the message it got a different hash.

jferrant (approved):

PR #387, opened 2023-05-30
Signer api warp

jferrant:
This is very much an initial sort of messing around to see if I could set up a database that was compatible with tokio for storing registered keys and some signer data and setting up some corresponding API endpoints for modifying and getting this info, making sure to support pagination in the process. I have written some basic unit tests for all db calls and routes as well. 

Next steps would include adding authentication, updating the database info to store the relevant info, add configuration options for the signer.

Closes https://github.com/Trust-Machines/core-eng/issues/410


stacks-signer-auto-api/src/db.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    pool.get()
+        .map_err(Error::from)?
+        .execute(sql_schema(), NO_PARAMS)?;
+    Ok(pool)
+}
+
+pub async fn add_signer(
+    signer: Signer,
+    pool: Arc<ConnectionPool>,
```
netrome, line 97: Why do we need to Arc the connection pool? It already uses an Arc internally and is cheap to clone.
https://docs.rs/r2d2/0.8.10/src/r2d2/lib.rs.html#317-319

stacks-signer-api/src/main.rs
```diff
@@ -0,0 +1,19 @@
+use stacks_signer_api::{db, routes};
+use std::sync::Arc;
+use warp::Filter;
+
+#[tokio::main]
+async fn main() {
+    // Initialize the connection pool
+    let pool = Arc::new(db::init_pool::<String>(None).expect("Failed to initialize pool."));
+
```
netrome, line 9: It would be nice to initiate tracing here to get some logs from warp.

stacks-signer-auto-api/src/db.rs
jferrant: Oh its def not needed. This was a search replace when I was using rusqlite directly that I just completely didn't think about after the fact.

stacks-signer-api/src/db/keys.rs
```diff
@@ -0,0 +1,308 @@
...
+
+const fn sql_count_keys_by_id() -> &'static str {
+    r#"
+    SELECT COUNT(*) FROM keys WHERE signer_id = ?1 AND user_id = ?2
+    "#
+}
+
+pub async fn add_key(key: Key, pool: SqlitePool) -> Result<impl warp::Reply, warp::Rejection> {
+    // First make sure we have an existing signer id;
+    let count: i64 = sqlx::query(sql_count_keys_by_id())
```
netrome, line 45: Future enhancements: The real magic of sqlx (compile-time verified queries etc.) comes when using the macros instead of the functions. That would require some tweaking here to get working properly (such as enabling "offline mode") so no need to bother with that here. 

I created this follow-up ticket to do the enhancements: https://github.com/Trust-Machines/core-eng/issues/408

Also, I didn't find any ticket linked in this PR. Where is the signer work tracked? Should I add any particular label to the follow-up ticket?

stacks-signer-api/src/db/keys.rs
```diff
@@ -0,0 +1,308 @@
...
+    Ok(warp::reply::with_status(
+        warp::reply::json(&serde_json::json!({ "status": "added" })),
+        http::StatusCode::CREATED,
+    ))
+}
+
+pub async fn delete_keys_by_id(
+    signer_id: i64,
+    user_id: i64,
+    pool: SqlitePool,
```
netrome, line 79: Although an Arc is cheap to clone a reference is still cheaper. There's no reason for this function to own the pool variable. We can save some reference count mutations and a few `.clone()` calls by taking a reference here (and in the other functions) instead.

```suggestion
    pool: &SqlitePool,
```

netrome (approved):
Good stuff. Just a minor comment and a follow-up suggestion.

stacks-signer-api/src/db/keys.rs
```diff
@@ -0,0 +1,308 @@
...
+        .fetch_all(&pool)
+        .await
+        .map_err(Error::from)?
+        .iter()
+        .map(|row: &sqlx::sqlite::SqliteRow| row.get(0))
+        .collect();
+    let page = query.page.unwrap_or(1);
+    let limit = query.limit.unwrap_or(usize::MAX);
+
+    let start_index = usize::min((page - 1) * limit, keys.len());
```
stjepangolemac, line 139: This might be a bit more idiomatic if it was:

```
keys.len().min((page - 1) * limit);
```

Similar for the cases below.

stacks-signer-api/src/db/keys.rs
```diff
@@ -0,0 +1,308 @@
...
+    use warp::Reply;
+
+    async fn init_db() -> SqlitePool {
+        init_pool(None)
+            .await
+            .expect("Failed to initialize a new database pool.")
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    #[ntest::timeout(1000)]
```
stjepangolemac, line 163: IMO specifying timeouts could be an anti-pattern as tests should usually run fast and be deterministic. It's okay if we want to keep those only until we're certain the tests are functioning properly.

It's also okay to leave the timeouts specified if that's what you prefer.

stacks-signer-api/src/routes/signers.rs
```diff
@@ -0,0 +1,205 @@
...
+use warp::Filter;
+
+#[derive(Debug, Deserialize, Clone)]
+pub struct SignerQuery {
+    pub status: Option<Status>,
+    pub page: Option<usize>,
+    pub limit: Option<usize>,
+}
+
+fn json_body_signer() -> impl Filter<Extract = (Signer,), Error = warp::Rejection> + Clone {
```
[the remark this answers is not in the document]
jferrant: You are right. It is sort of duplicated but not directly. I probably can turn it into a generic though and pass the Key and SIgner as the generic type...will update.

stacks-signer-api/src/db/keys.rs
jferrant: These all take a SqlitePool because I could not  satisfy the warp api/lifetimes without passing pool.clone() which then forces the subsequent function signature to take the object rather than a reference. I can try again, but I have my doubts that it is possible.

stacks-signer-api/src/db/keys.rs
jferrant: Because of sqlx there are blocking calls being made by these tests. This will not work unless I run with multi-thread (Removing it will result in
```---- db::keys::tests::test_add_key stdout ----
thread 'db::keys::tests::test_add_key' panicked at 'can call blocking only when running on the multi-threaded runtime', /Users/jferrant/.cargo/registry/src/github.com-1ecc6299db9ec823/sqlx-core-0.5.1/src/sqlite/connection/establish.rs:60:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace```
I added the timeout since, it being a blocking call, I didn't want it to potentially be blocked forever and added the timeout per @netrome 's suggestion

File: stacks-signer-api/src/db/keys.rs [hunk 8]
```diff
@@ -0,0 +1,308 @@
...
+    use warp::http::StatusCode;
+    use warp::Reply;
+
+    async fn init_db() -> SqlitePool {
+        init_pool(None)
+            .await
+            .expect("Failed to initialize a new database pool.")
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
```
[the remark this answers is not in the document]
jferrant: Because of sqlx there are blocking calls being made by these tests. This will not work unless I run with multi-thread.  (Removing it will result in
``---- db::keys::tests::test_add_key stdout ----
thread 'db::keys::tests::test_add_key' panicked at 'can call blocking only when running on the multi-threaded runtime', /Users/jferrant/.cargo/registry/src/github.com-1ecc6299db9ec823/sqlx-core-0.5.1/src/sqlite/connection/establish.rs:60:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace``

stacks-signer-api/src/db/keys.rs
jferrant: > 

I created a new tag: `sbtc-signer` 
I think they also should be tagged with `sbtc-mini`. 

See https://github.com/Trust-Machines/core-eng/issues/410

stacks-signer-api/src/db/keys.rs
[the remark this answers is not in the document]
stjepangolemac: Ah okay. Thanks for the info 👍

stacks-signer-api/src/db/keys.rs
netrome: Yep, this is to ensure deadlocks does not hang the test runner for long periods which it would otherwise do. Any test case that comes close to 1s of runtime is broken in one way or another, and there's no need to wait any longer imo.

stacks-signer-api/src/db/mod.rs
```diff
@@ -0,0 +1,49 @@
...
+        Some(path) => SqlitePool::connect(path).await?,
+        None => SqlitePool::connect("sqlite::memory:").await?,
+    };
+    sqlx::query(sql_schema_signers()).execute(&pool).await?;
+
+    sqlx::query(sql_schema_keys()).execute(&pool).await?;
+
+    Ok(pool)
+}
+const fn sql_schema_signers() -> &'static str {
```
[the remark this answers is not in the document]
netrome: Also mentioned as a potential follow-up in https://github.com/Trust-Machines/core-eng/issues/408

stacks-signer-api/src/db/keys.rs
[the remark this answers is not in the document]
netrome: "worker_threads = 1" still works though if we want to do single-threaded execution.

PR #399, opened 2023-05-31
Set the coordinator data before setting the bitcoin wallet address

jferrant:
Closes: https://github.com/Trust-Machines/core-eng/issues/398

we have success!
Setting the coordinator data: https://explorer.hiro.so/txid/0x1e93aae744144dbb2611c64ee6a53ea2de926665091cda5e7381dc52cf161292?chain=mainnet
Setting the bitcoin wallet public key: https://explorer.hiro.so/txid/0x64be18d77557d6577f9e6782b3f5dbe2924aaed848af2a384ff96a4389b58453?chain=mainnet


stacks-coordinator/src/stacks_node/client.rs
```diff
@@ -221,8 +223,13 @@ impl StacksNode for NodeClient {
...
     }
 
-    fn next_nonce(&self, address: &StacksAddress) -> Result<u64, StacksNodeError> {
+    fn next_nonce(&mut self, address: &StacksAddress) -> Result<u64, StacksNodeError> {
         debug!("Retrieving next nonce...");
+        if let Some(nonce) = self.next_nonce {
+            let next_nonce = nonce.wrapping_add(1);
+            self.next_nonce = Some(next_nonce);
+            return Ok(next_nonce);
+        }
```
netrome, line 232: Hmm... this feels brittle if we restart or do anything else. I guess this is fine for now, but I would be more comfortable with a solution where we await confirmation from the chain from one transaction before we broadcast the next one.

stacks-coordinator/src/stacks_wallet.rs
```diff
@@ -192,17 +198,43 @@ impl StacksWalletTrait for StacksWallet {
...
+        let key = Value::Sequence(SequenceData::Buffer(BuffData {
+            data: public_key.to_bytes_compressed(),
+        }));
+        let data = TupleData::from_data(vec![
+            (ClarityName::from("addr"), principal),
+            (ClarityName::from("key"), key),
+        ])
+        .map_err(Error::from)?;
+        let function_args = vec![data.into()];
+        // (data {addr: principal, key: (buff 33)}))
```
netrome, line 226: Nit: This is declared just a few lines above. No need to comment it here. 
```suggestion
```

netrome (approved):
lgtm :+1:

PR #402, opened 2023-06-01
Do not broadcast set coordinator data tx if already correctly set in the contract

jferrant:
Closes https://github.com/Trust-Machines/core-eng/issues/401


sergey-shandar (approved):

xoloki (approved):
LGTM

PR #405, opened 2023-06-01
Load the wallet address using importdescriptors

jferrant:
Closes https://github.com/Trust-Machines/core-eng/issues/397


stacks-coordinator/src/bitcoin_node.rs
```diff
@@ -177,13 +184,19 @@ impl LocalhostBitcoinNode {
     }
 
     fn import_address(&self, address: &BitcoinAddress) -> Result<(), Error> {
-        let address = address.to_string();
         debug!("Importing address {}...", address);
+        // Create a descriptor using a Bech32 (segwit) Pay-to-Witness Public Key Hash (P2WPKH) address.
+        let desc = format!("wpkh({})", address);
```
netrome, line 189: I suspect this is incorrect. We're using a taproot address and not a p2wpkh address. I guess we don't have any transactions on-chain that we have verified this with right? I.e. our only verification so far is that the call succeeds?

netrome: Update: It seems like we should be good if we simply do it like this
```suggestion
        // Create a descriptor using a Bech32 (segwit) Pay-to-Taproot (P2TR) address.
        let desc = format!("tr({})", address);
```

netrome (approved):

stacks-coordinator/src/bitcoin_node.rs
netrome: I committed a few shell scripts I used to play around with this and understand these RPC calls better. They are under `/misc`.

jferrant: Feel free to commit this change.

netrome: Oh I thought I already did that :sweat_smile: Done now :heavy_check_mark:

stjepangolemac (approved):

stacks-coordinator/src/bitcoin_node.rs
```diff
@@ -70,12 +70,14 @@ pub struct LocalhostBitcoinNode {
...
 impl BitcoinNode for LocalhostBitcoinNode {
     fn broadcast_transaction(&self, tx: &BitcoinTransaction) -> Result<Txid, Error> {
-        let mut writer = Cursor::new(vec![]);
-        tx.consensus_encode(&mut writer)?;
-        let raw_tx = writer.into_inner().to_hex();
+        let mut tx_bytes: Vec<u8> = vec![];
+        tx.consensus_encode(&mut tx_bytes)?;
+        let raw_tx = hex::encode(&tx_bytes);
+
+        //let raw_tx = writer.into_inner().to_hex();
```
netrome, line 73: Nit: Remove (I'll fix)
```suggestion
```

stacks-coordinator/tests/local_bitcoin_node.rs
```diff
@@ -1,30 +1,130 @@
...
+    let deposit_wallet_public_key_secp =
+        bitcoin::secp256k1::PublicKey::from_slice(&deposit_wallet_public_key.to_bytes()).unwrap();
+    let deposit_wallet_public_key_xonly = XOnlyPublicKey::from(deposit_wallet_public_key_secp);
+    let deposit_wallet_public_key_tweaked =
+        TweakedPublicKey::dangerous_assume_tweaked(deposit_wallet_public_key_xonly);
+    println!(
+        "build address with shared public key {} tweaked {:?}",
+        deposit_wallet_public_key_secp, deposit_wallet_public_key_tweaked
+    );
+    let taproot = Script::new_v1_p2tr_tweaked(deposit_wallet_public_key_tweaked);
```
netrome: Ahhh! This is the key piece we missed before introducing the big changes in the last two commits. We were applying the tap tweak to the address we mined to, but not when we produced the signature. We would have to either apply the tap tweak while signing, or do like we do here and not apply the tap tweak when generating the address (which is how we do it with FROST).

This simple diff actually fixes the broken `should_broadcast_transaction` failure in `1eae910f0a07a7cdee6d72d6aebbd7cdd3a71dd5`:

```
diff --git a/test-utils/src/lib.rs b/test-utils/src/lib.rs
index ad095f3..8b11849 100644
--- a/test-utils/src/lib.rs
+++ b/test-utils/src/lib.rs
@@ -1,4 +1,5 @@
 use bitcoin::psbt::Prevouts;
+use bitcoin::schnorr::TweakedPublicKey;
 use bitcoin::secp256k1::{All, Message, Secp256k1, SecretKey};
 use bitcoin::util::sighash::SighashCache;
 use bitcoin::{
@@ -230,8 +231,9 @@ pub fn generate_wallet() -> (SecretKey, PrivateKey, PublicKey, XOnlyPublicKey, A
     let private_key = PrivateKey::new(secret_key, Network::Regtest);
     let public_key = PublicKey::from_private_key(&secp, &private_key);
     let xonly_public_key = keypair.x_only_public_key().0;
-    let secp = Secp256k1::verification_only();
-    let address = bitcoin::Address::p2tr(&secp, xonly_public_key, None, Network::Regtest);
+
+    let assume_tweaked_public_key = TweakedPublicKey::dangerous_assume_tweaked(xonly_public_key);
+    let address = bitcoin::Address::p2tr_tweaked(assume_tweaked_public_key, Network::Regtest);
 
     //let address = Address::p2wpkh(&public_key, Network::Regtest).unwrap();
 
@@ -295,7 +297,4 @@ pub fn sign_transaction(
     tx.input[0]
         .witness
         .push(bitcoin::psbt::serialize::Serialize::serialize(&schnorr_sig));
-    tx.input[0]
-        .witness
-        .push(bitcoin::psbt::serialize::Serialize::serialize(public_key));
 }
```

That being said, I think it's good that we're reusing the signer code here instead. I just wanted to understand why the previous implementation failed and this one worked. Now I'm happy :relieved:

PR #414, opened 2023-06-07
feat: Make relay port configurable

stjepangolemac:
Port is now an argument and defaults to the previously hardcoded port.


netrome (approved):
LGTM, just one minor suggestion.

relay-server/src/bin/relay-server.rs
```diff
@@ -1,6 +1,13 @@
...
+    #[clap(short, long, default_value_t = 9776)]
+    port: u16,
+}
+
 fn main() {
-    let addr = "127.0.0.1:9776";
-    Server::run(addr);
+    let args = Cli::parse();
+    Server::run(&format!("127.0.0.1:{}", args.port));
 }
```
netrome, line 13: Also, I'd swap the order of these functions so main comes first.

PR #430, opened 2023-06-13
Make logs of wallet error and peg transactions lookup better

stjepangolemac:


netrome (approved):
Wow! Such beautiful! Many println!

PR #437, opened 2023-06-13
Add and fix logging in block height polling

jferrant:
Closes https://github.com/Trust-Machines/core-eng/issues/435

This actually was not a bug in the range. It is intended functionality that coordinator default to the tip of the chain and process that block. It then will not process anything after this is processed until new blocks show up. however the logging is confusing so I make sure to exit before logging if we have already processed the current tip. I also added some logging to the poll function as it can appear as if the coordinator is hanging while it waits for the block height to advance.


soju-drinker (approved):

netrome (approved):

stjepangolemac (approved):

stjepangolemac:
Don't mind me merging this, thanks for the fix 🙏

PR #453, opened 2023-06-15
Increment the next nonce when a broadcast fails due to ConflictingNonceInMempool

jferrant:


stjepangolemac (commented):
I suggest us to be very careful with unbounded loops. Maybe we could make it so it tries N times and then crashes?

Also, the looping logic seems idential between transactions, maybe it needs to be deduplicated and put somewhere once.

stacks-coordinator/src/coordinator.rs
```diff
@@ -122,32 +122,67 @@ trait CoordinatorHelpers: Coordinator {
...
+                Ok(_) => return Ok(()),
+            }
+        }
     }
 
     fn peg_out(&mut self, op: stacks_node::PegOutRequestOp) -> Result<()> {
         // Retrieve the nonce from the stacks node using the sBTC wallet address
         let address = *self.fee_wallet().stacks().address();
-        let nonce = self.stacks_node_mut().next_nonce(&address)?;
+        let mut nonce = self.stacks_node_mut().next_nonce(&address)?;
```
netrome, line 158: Small simplification suggestion part 1: We don't need this variable

```suggestion
```

stacks-coordinator/src/coordinator.rs
```diff
@@ -122,32 +122,67 @@ trait CoordinatorHelpers: Coordinator {
...
+                .stacks()
+                .build_burn_transaction(&op, nonce)?;
+
+            // Broadcast the resulting sBTC transaction to the stacks node
+            match self.stacks_node().broadcast_transaction(&burn_tx) {
+                Err(StacksNodeError::BroadcastError(BroadcastError::ConflictingNonceInMempool)) => {
+                    warn!("Burn transaction rejected by stacks node due to conflicting nonce in mempool. Stacks node may be falling behind!");
+                    warn!("Incrementing nonce and retrying...");
+                    nonce = self.stacks_node_mut().next_nonce(&address)?;
+                }
```
netrome, line 173: Part 2: Compute the nonce first thing in the loop
```suggestion
        loop {
            // Note: This mutates `self.stacks_node` each iteration
            let nonce = self.stacks_node_mut().next_nonce(&address)?;
            
            // Build a burn transaction using the withdrawal request op and calculated nonce
            let burn_tx = self
                .fee_wallet()
                .stacks()
                .build_burn_transaction(&op, nonce)?;

            // Broadcast the resulting sBTC transaction to the stacks node
            match self.stacks_node().broadcast_transaction(&burn_tx) {
                Err(StacksNodeError::BroadcastError(BroadcastError::ConflictingNonceInMempool)) => {
                    warn!("Burn transaction rejected by stacks node due to conflicting nonce in mempool. Stacks node may be falling behind!");
                    warn!("Incrementing nonce and retrying...");
                }
```

netrome (approved):
Just a suggestion, otherwise lgtm

jferrant:
> I suggest us to be very careful with unbounded loops. Maybe we could make it so it tries N times and then crashes?
> 
> Also, the looping logic seems idential between transactions, maybe it needs to be deduplicated and put somewhere once.

The looping logic is not identical because they call separate functions unfortunately. But will clean up. :D

stacks-coordinator/src/coordinator.rs
jferrant: I cannot do this. If I did, I would increment the nonce in the fee to low case which is incorrect.

stacks-coordinator/src/coordinator.rs
jferrant: See part 1 response. This is not possible because there is the failure case that occurs when the fee is too low which should not increment the nonce in this case.

stacks-coordinator/src/stacks_wallet.rs
```diff
@@ -142,38 +142,48 @@ impl StacksWallet {
...
-    fn build_mint_transaction(
+pub trait TransactionBuilder {
+    fn build_transaction(
+        &self,
+        wallet: &StacksWallet,
+        nonce: u64,
+    ) -> Result<StacksTransaction, PegWalletError>;
+}
+
+impl TransactionBuilder for PegInOp {
```
netrome, line 153: I find this a bit confusing to read. I get the intention to break out shared logic, but `TransactionBuilder` sounds weird. For one, I'd expect the transaction to be the bitcoin transaction associated with the ops. Not the stacks transaction that the coordinator creates following these ops.

I understand that there's a 1:1 relation between the PegInOp -> mint tx and PegOutRequestOp -> burn tx that we want to leverage. Perhaps another less confusing name for this trait would be `SubsequentStacksTransaction`. 

Another more flexible (but perhaps over-engineering) solution would be to create concrete types for the Mint and Burn transactions, implementing a common trait and some type of conversion trait. Something like:

```rust
struct Mint {...};
struct Burn {...};

impl StacksTransaction for Mint;
impl StacksTransaction for Burn;

impl BuildTransaction<Mint> for PegInOp;
impl BuildTransaction<Burn> for PegOutRequestOp;

impl StacksWalletTrait for StacksWallet {
    fn build_transaction<T: StacksTransaction>(
        &self,
        op: impl BuildTransaction<Target = T>,
        nonce: u64,
    ) 
    
pub trait StacksTransaction;
pub trait BuildTransaction{
    type Target = StacksTransaction;
}
```

jferrant: I think the concrete types to enable a one op to many transaction types from a stackswallet may be useful in the future, but won't be hard to pull out if required when the time comes and may just overcomplicate the code for no reason at the moment. I will keep it as is but with a rename to "BuildStacksTransaction" to be explicit about the type of transaction being generated from a provided Op. :D

stjepangolemac (approved):

PR #470, opened 2023-06-21
fix: Remove fulfillment_fee from withdrawal data in sbtc-cli

### LINKED ISSUE #469: sBTC CLI withdrawal transactions data is wrong
Currently the withdrawal transactions constructed with the sBTC CLI append the fulfillment fee to the data. They should not do this.

### Definition of done
The fulfillment fee is not appended to the data in the withdrawal transactions constructed by the sBTC CLI.

Note: We can also omit zero padding the end of the data, which I will do as well.

netrome:
Closes #469


sbtc-cli/src/main.rs
```diff
@@ -298,9 +292,6 @@ fn withdrawal_data(
         .chain(amount.to_be_bytes())
         .chain(once(recovery_id.to_i32() as u8))
         .chain(signature)
-        .chain(repeat(0))
```
jferrant, line 301: Does this mean test vectors at https://github.com/Trust-Machines/stacks-sbtc/blob/dece0a6474450a4fd45aae9d181a50b3285277d8/test-vectors/src/peg_in.rs#L165 and at https://github.com/Trust-Machines/stacks-sbtc/blob/dece0a6474450a4fd45aae9d181a50b3285277d8/test-vectors/src/peg_out.rs#L195 are wrong? If so, I will make a ticket to fix them (this is where I got the code from).

netrome: The test vectors are correct. The padding is optional, so I just chose to omit it here to save the block space. The error was the fulfillment fee, which should not be part of the data.

jferrant: No at the test vectors they add the fee.

netrome: That's the `fee_subsidy` which is only added in commit-reveal to subsidize the fee of the Reveal transaction, which is not the same as the `fulfillment_fee` which is a separate output to subsidize the Fulfillment transaction.

That's why the `fee_subsidy` is optional in the test vectors. It's only added in commit-reveal, and ignored in OP_RETURN.

I understand the confusion. To add to this confusion, there was discussions on omitting the fulfillment fee in favor of making the fee subsidy cover both the fulfillment and reveal transaction fees. But this is only particular for the withdrawal request commit transaction and does not apply to any other transaction. I am not aware of any decision on this though, and to keep things simple I prefer to save it as a future optimization.

Reference: https://github.com/Trust-Machines/stacks-sbtc/blob/main/sbtc-planning/commit-reveal-ops.md#the-commit-reveal-operation-format

stjepangolemac (approved):

PR #501, opened 2023-06-29
Update README.md with Clarity Test Badge

jo-tm:
I am testing if this Clarity Test Badge is working cc/ @sergey-shandar


jo-tm:
Seems to be working fine:

image

..

image

PR #504, opened 2023-06-30
Build both withdrawal sBTC and BTC transactions before attempting broadcast

jferrant:
This prevents a half completed state to occur when a fulfilled tx is incomplete. Otherwise we will continually create sBTC transactions but never the BTC one.


stjepangolemac (approved):
Looks good. Depending on which one is of higher importance maybe it's worth broadcasting that one first.

PR #510, opened 2023-07-05
Fulfill withdrawal requests by adding missing withdrawal output and fixing dust amount

jferrant:
Need to incorporate @xoloki's fix to setting the aggregate public key (signing fulfill_tx fails).
I also am unsure about the address being tored within the wallet. I believe this should be a TweakedPublicKey however changing this throughout the code (as well as generate_wallet in the tests) causes both list_unspent and load_wallet tests to fail. Feedback?


stacks-coordinator/src/bitcoin_node.rs
```diff
@@ -220,7 +223,10 @@ impl LocalhostBitcoinNode {
 
         // Create a descriptor using a Bech32 (segwit) Pay-to-Taproot (P2TR) address.
         let desc = {
-            let pub_hex = hex::encode(wallet.x_only_pub_key().serialize());
+            let pub_hex = hex::encode(
+                TweakedPublicKey::dangerous_assume_tweaked(wallet.x_only_pub_key().clone())
```
jferrant, line 227: I am not 100% confident in this change, but from googling around and my limited understanding of p2tr...This descriptor expects a tweaked public key. HOWEVER, removing this change does not break the tests....Which is...concerning...

stacks-coordinator/src/bitcoin_wallet.rs
```diff
@@ -96,12 +91,14 @@ impl BitcoinWalletTrait for BitcoinWallet {
             "change_amount: {:?}, total_consumed: {:?}, op.amount: {:?}",
             change_amount, total_consumed, op.amount
         );
-        if change_amount >= DUST_UTXO_LIMIT {
-            let secp = Secp256k1::verification_only();
-            let script_pubkey = Script::new_v1_p2tr(&secp, self.public_key, None);
+        // Do not want to use Script::new_v1_p2tr because it will tweak our key when we don't want it to
+        let public_key_tweaked = TweakedPublicKey::dangerous_assume_tweaked(self.public_key);
```
jferrant, line 95: Removing this will result in an invalid script signature. The public key returned by the DKG round is already a tweaked key and should not be retweaked.

test-utils/src/lib.rs
```diff
@@ -251,7 +251,7 @@ pub fn generate_wallet(
     let public_key = PublicKey::from_private_key(&secp, &private_key);
     let xonly_public_key = keypair.x_only_public_key().0;
     let address = if is_taproot {
-        bitcoin::Address::p2tr(&secp, xonly_public_key, None, Network::Regtest)
+        Address::p2tr(&secp, xonly_public_key, None, Network::Regtest)
```
jferrant, line 254: This is one location where I am worried we should actually be doing p2tr_tweaked...

stacks-coordinator/src/bitcoin_wallet.rs
```diff
@@ -23,23 +21,20 @@ pub enum Error {
...
-    public_key: PublicKey,
+    public_key: XOnlyPublicKey,
 }
 
 impl BitcoinWallet {
-    pub fn new(public_key: PublicKey, network: Network) -> Self {
+    pub fn new(public_key: XOnlyPublicKey, network: Network) -> Self {
         let secp = Secp256k1::verification_only();
-        let address = bitcoin::Address::p2tr(&secp, public_key, None, network);
+        let address = Address::p2tr(&secp, public_key, None, network);
```
jferrant, line 30: This is one location where I am worried we should actually be doing p2tr_tweaked...However, if we do this as well as in the test where I commented, we will break list_unspent and should_load_wallet

stacks-coordinator/src/bitcoin_node.rs
xoloki: Here you're directly encoding the key into hex, so there's no reason to do use a tweaked pubkey.  You only need to do the dangerous call if you're creating a TweakedPublicKey, not bytes.  `TweakedPublicKey::dangerous_assume_tweaked` is a no-op.

test-utils/src/lib.rs
xoloki: As long as `p2tr` isn't expecting a TweakedPublicKey, the XOnlyKey should be fine.

netrome: Nice spotted. This needs to be `p2tr_tweaked` since `p2tr` will tweak the address given to it. The only reason the tests are currently not failing is since the `BitcoinNode::new()` implementation also uses `p2tr` resulting in a tweaked address. I went down the rabbit hole here and reproduced the failing test cases by switching `p2tr` to `p2tr-tweaked` in those places. After some debugging I noticed that the `tr(...)` descriptor we use in `LocalBitcoinNode::import_address` also applies the tap tweak. It would be more robust if this function used the `addr` descriptor instead.

Since this problem is orthogonal to what you're working on here, I'll submit this fix as a separate PR.

netrome (approved):
Thanks for bringing the `Address::p2tr(..)` problem to my attention, I'll submit a PR to fix it separately.

netrome:
I just pushed a rebased version of this branch `447-sbtc-alpha-withdrawal-requests-arent-fulfilled-rebase-1` since I need it for alpha testing. I did not want to mess up the original one, but feel free to hijack that rebase if it looks good.

stacks-coordinator/src/bitcoin_wallet.rs
netrome: Reading this again, we should do `p2tr_tweaked` because we don't want to tweak the key. Aaaargh this is confusing :sweat_smile:.

Anyway, I've changed this in https://github.com/Trust-Machines/stacks-sbtc/pull/531

netrome (approved):

stjepangolemac (approved):

netrome (approved):

PR #513, opened 2023-07-06
chore: remove protocol-caller parameter

MarvinJanssen:
We don't need this parameter in sub contracts. It has already been removed in other contracts but it remained in the token contract.


jo-tm:
@jo-tm @setzeus review this.

jo-tm (approved):

jo-tm:
@setzeus pls review.

friedger (approved):
Looks good!

PR #516, opened 2023-07-06
Add documentation and clarification to clarity-bitcoin

friedger:
This PR
* adds documentation to the bitcoin library
* replaces state-res with result and uses `(state (unwrap! result result)` instead of `match`
* adds a function `bool-list-of-len` to generate a list of true values
* uses stacks items instead of elements in describing stack items of witness data
* slices the iterator to the tree-depth
* removes `was-wtx-mined-internal`


friedger:
Unit tests fail in this repo but pass in https://github.com/friedger/clarity-bitcoin/pull/14

jo-tm:
please send a PR for the good branch @friedger

friedger:
I would like to use the same code in both repos (or link them or something like that) so that we have one version of the bitcoin library. I don't have a good idea how to do that.

friedger:
@FriendsFerdinand Thank you for fixing this! It would have taken me ages to solve it.

FriendsFerdinand (approved):

MarvinJanssen (approved):
LGTM. We can optimise the lib later.

PR #527, opened 2023-07-07
Add sBTC wif to withdrawal args

### LINKED ISSUE #532: Support withdrawing from arbitrary Stacks addresses in sBTC-cli
Currently, the `sbtc-cli` assumes the same key to be both used to sign the sBTC withdrawal transaction and sign the withdrawal payload. This means that you have to consume a BTC utxo from the BTC address that corresponds to the Stacks address holding the sBTC to initiate a withdrawal.

This is a severe limitation only intended for alpha. However, we have learned that the Hiro Wallet does not use the same keys for the Stacks addresses and BTC addresses - meaning we are not able to use this simplification even for alpha.

Hence, we should support using different keys for signing the withdrawal transaction and signing the payload (which also identifies the sBTC sender).

### Definition of done
The `sbtc-cli` does not require the sBTC sender key to be the same key that signs the withdrawal BTC transaction.

jferrant:
Closes #532


netrome (commented):
~What is the problem with using the existing `wif` argument for the withdrawal? This change seems to introduce a new argument which does exactly the same thing as the `wif` argument used to do, but only for withdrawals.~

Nevermind, the `wif` is also used to sign the BTC transaction and it makes sense to not require the sBTC sender and the BTC transaction signer to be the same.

netrome (approved):

netrome:
I'm taking the liberty to merge this PR now.

PR #537, opened 2023-07-12
feat: alpha coordinator persisting dkg public shares

### LINKED ISSUE #536: [sbtc-alpha] coordinator needs to persist DKG polynomials or derived values
**Describe the bug**
Current `frost-coordinator` expects that the DKG polynomials are available during the signing process.  They are used to generate the aggregate public key, and public keys for each key share; the latter keys are needed to validate signature shares from the signers.  

This is not a problem during unit or integration testing, since in those contexts we do both DKG and sign operations. However, in a production environment there are no guarantees that coordinator will never go down.

The sBTC alpha contract currently stores the aggregate public key, but it does not store the share public keys, or the polynomials themselves.  The amount of storage needed depends on what we choose to store:

For `N` keys and threshold `T` (where `T < N`), storing all polynomials will be `O(N*T)`.  Storing only the aggregate polynomial will be `O(T)`, and storing the share public keys will be `O(N)`.  So if the only consideration is space required, then we should store the aggregate polynomial.

However, if we store the aggregate polynomial then we will have to calculate the share public keys during every signing operation.  This roughly doubles the time required to sign.  So the best tradeoff appears to be storing the share public keys.

Since we are storing the aggregate public key in the contract, it makes the most sense to store the share public keys there as well.

**To Reproduce**
1. Run DKG using `stacks-coordinator` and `stacks-signer`.  This will persist the aggregate public key in the contract.
2. Restart `stacks-coordinator`, then try to do a signing round.  The aggregate public key will load from the contract, but not the polynomials.
3. Signing will fail

**Expected behavior**
Signing should work after a `stacks-coordinator` restart.

netrome:
closes #536


jferrant (approved):

stacks-coordinator/src/coordinator.rs
```diff
@@ -365,23 +372,70 @@ fn create_frost_coordinator(
...
 ) -> Result<XOnlyPublicKey> {
     debug!("Retrieving bitcoin wallet public key from sBTC contract...");
     if let Some(xonly_pubkey) = stacks_node.bitcoin_wallet_public_key(address)? {
-        // We have to set the frost_coordinator aggregate key
-        frost_coordinator.set_aggregate_public_key(
-            Point::lift_x(&Element::from(xonly_pubkey.serialize()))
-                .map_err(|e| Error::PointError(format!("{:?}", e)))?,
-        );
+        if let Some(data_directory) = data_directory {
+            frost_coordinator.set_dkg_public_shares(read_dkg_public_shares(data_directory)?);
```
jferrant, line 427: Ah this is a bug I. missed. We still need to set the aggregate_public_key...

PR #539, opened 2023-07-12
set the aggregate public key properly

jferrant:


netrome (approved):
Thank you for spotting this :pray:

PR #557, opened 2023-07-13
Add state to processed ops and trim the table

jferrant:
I didn't have internet for a majority of the day, so just spent some time looking over the code and seeing what sort of improved state I could easily add.

This is not entirely necessary, but now it will poll for withdrawals and deposits and save its last processed state. It should help with continually broadcasting sBTC transactions for a half processed transaction if the coordinator crashes. It will be smart enough to pick up where it left off to some extent now. Still not perfect by any means but I think its a worthwhile change. Feel free to suggest improvements on it.

I trim the table of completed transactions, however, I could leave them and add logic to not overwrite completed transactions with new ones if we set the blockheight back in time, but I feel like we would expect these to be forcibly reprocessed if we are setting the blockheight back...so I think trimming the table makes more sense. (The case of the coordinator processing most but not all the transactions at some block height seems unlikely and would be the only scenario I can think of where we would want to go back in time and process some but not all the transactions at a given block height...We currently poll all peg ins and all peg outs before attempting processing so they would all be already stored in the database with status::NEW for a given block height)

### DIFF
--- a/stacks-coordinator/src/bitcoin_wallet.rs
+++ b/stacks-coordinator/src/bitcoin_wallet.rs
@@ -266,7 +266,8 @@ mod tests {
         assert_eq!(btc_tx.input.len(), 7);
         assert_eq!(btc_tx.output.len(), 3); // We have change!
         assert_eq!(btc_tx.output[0].value, 0);
-        assert_eq!(btc_tx.output[1].value, 10000);
+        assert_eq!(btc_tx.output[1].value, amount);
+        assert_eq!(btc_tx.output[2].value, 10000);
     }
 
     #[test]
--- a/stacks-coordinator/src/coordinator.rs
+++ b/stacks-coordinator/src/coordinator.rs
@@ -6,7 +6,10 @@ use bitcoin::{
     },
     SchnorrSighashType, XOnlyPublicKey,
 };
-use blockstack_lib::{types::chainstate::StacksAddress, util::secp256k1::Secp256k1PublicKey};
+use blockstack_lib::{
+    chainstate::stacks::StacksTransaction, types::chainstate::StacksAddress,
+    util::secp256k1::Secp256k1PublicKey,
+};
 use frost_coordinator::{
     coordinator::Error as FrostCoordinatorError, create_coordinator, create_coordinator_from_path,
 };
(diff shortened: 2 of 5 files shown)


stacks-coordinator/src/bitcoin_wallet.rs
```diff
@@ -266,7 +266,8 @@ mod tests {
         assert_eq!(btc_tx.input.len(), 7);
         assert_eq!(btc_tx.output.len(), 3); // We have change!
         assert_eq!(btc_tx.output[0].value, 0);
-        assert_eq!(btc_tx.output[1].value, 10000);
```
jferrant, line 269: This is an unrelated change to fix tests broken by the last merge. I can pull this out later, but just leaving it here until I get eyes on this PR.

stacks-coordinator/src/coordinator.rs
```diff
@@ -126,44 +126,110 @@ pub trait Coordinator: Sized {
...
     fn process_queue(&mut self) -> Result<()> {
-        match self.peg_queue().sbtc_op()? {
-            Some(SbtcOp::PegIn(op)) => {
-                debug!("Processing peg in request: {:?}", op);
-                self.peg_in(op)
-            }
-            Some(SbtcOp::PegOutRequest(op)) => {
-                debug!("Processing peg out request: {:?}", op);
-                self.peg_out(op)
+        loop {
```
jferrant, line 129: This change I think is a good one to have even if we don't have the added status...before we were polling everything but only processing one sbtc op per poll call...I think we should poll everything, process everything, then poll again, etc.

stacks-coordinator/src/coordinator.rs
```diff
@@ -126,44 +126,110 @@ pub trait Coordinator: Sized {
...
+                    );
+                }
+                Status::Complete => {
+                    // We're done
+                    self.peg_queue()
+                        .update_status(&op.txid, &op.burn_header_hash, status)?;
+                    return Ok(());
+                }
+            }
+            self.peg_queue()
```
jferrant, line 230: In theory we could crash right before this call when the broadcast function has already gone through, but ....not sure how we could ever prevent that XD

stacks-coordinator/src/coordinator.rs
netrome: Yeah that makes sense.

netrome (changes requested):
It's not obvious to me how the addition of the peg queue status improves anything in terms of processing policy, and I'm afraid the introduction of the new complexity might bring on more bugs.

I recommend holding this change until we know if it's worth spending more time on the coordinator.

stjepangolemac (commented):
I see this PR as 3 different types of changes:

1. Smaller miscellaneous fixes
2. Batching when processing operations
3. Additional statuses when processing

First type I'd pull out into a separate PR and review / merge quickly, no issues there. 

Second one doesn't make a material difference for the execution so I don't have any strong preferences there. Although, if we want to get closer to a state machine, a single operation / event is closer to the ideal.

When it comes to additional bookkeeping of statuses when processing, I think that's something we need but it's very hard to even write a spec for it. There's a good chance we might introduce additional complexity and bugs without being confident that we've covered all the cases. I think we first need to find out if it's worth to spend more time on the coordinator, and if it is then first become confident about the spec, then the implementation.

jferrant:
> It's not obvious to me how the addition of the peg queue status improves anything in terms of processing policy, and I'm afraid the introduction of the new complexity might bring on more bugs.
> 
> I recommend holding this change until we know if it's worth spending more time on the coordinator.

The only benefit is if for some reason coordinator crashes after having polled the node and partially finished processing an sBTC op...it will pick up where it left off for that op and not rebroadcast an sBTC transaction if it's already done so

PR #576, opened 2023-07-17
[sbtc-cli] Move commands in their respective files

### LINKED ISSUE #575: Break out sBTC CLI subcommands into spearate files
Currently, the sBTC CLI has grown to become a quite large single `main.rs` file. It would be easier to read the code if the different commands were broken out to their own separate modules. This would prepare some of the functionality for reuse as well.

### Definition of done
The functionality for the sBTC CLI subcommands are broken out to separate modules.

stjepangolemac:
Closes #575


sbtc-cli/src/commands/broadcast.rs
```diff
@@ -0,0 +1,26 @@
+use anyhow::anyhow;
+use bdk::blockchain::Blockchain;
+use bitcoin::{psbt::serialize::Deserialize, Network, Transaction};
+use clap::Parser;
+
+use crate::commands::utils::init_blockchain;
```
netrome, line 6: Nit: I'm not a fan of importing functions. Can you keep the `utils` namepsace here, so instead just do `use crate::commands::utils`?

netrome (approved):
LGTM. Just one minor nit.

sbtc-cli/src/commands/deposit.rs
```diff
@@ -0,0 +1,103 @@
...
+use anyhow::anyhow;
+use bdk::{database::MemoryDatabase, SignOptions, Wallet};
+use bitcoin::{
+    psbt::{serialize::Serialize, PartiallySignedTransaction},
+    Address as BitcoinAddress, Network, PrivateKey,
+};
+use blockstack_lib::types::{chainstate::StacksAddress, Address};
+use clap::Parser;
+
+use crate::commands::utils::{build_op_return_script, magic_bytes, reorder_outputs, setup_wallet};
```
netrome, line 12: Same here. Could you just import `crate::commands::utils` so it's clear for a reader that these functions are not defined in this module.

netrome (commented):
One comment went missing in my previous review. Weird :thinking:

Issue #562, opened 2023-07-14
[sbtc-mini] sbtc-stacking-pool / signer-pre-register-test should be read only


radicleart:
**signer-pre-register-test**

`sbtc-stacking-pool.signer-pre-register-test` does not change state - should be read only.

setzeus:
Copy @radicleart . This was a placeholder function for developing a generation script on my end. Will remove in next PR.

jo-tm:
PR #578 will fix this.

friedger:
fixed by #578

Issue #548, opened 2023-07-13
[stacking-pool] vote-for-threshold-wallet-candidate does not extend stacking


friedger:
**Describe the bug**
vote-for-threshold-wallet-candidate only calls stack-aggregation-commit and does not call delegate-stack-extend

**To Reproduce**
Inspect function https://github.com/Trust-Machines/stacks-sbtc/blob/a9157761c9e6f7be76e76a38c72aa36a822086f2/sbtc-mini/contracts/sbtc-stacking-pool.clar#L437

**Expected behavior**
voting means that the user locks their stx for the new wallet, therefore the stacking must be extended.

stack-aggregation-commit call ensures that the wallet eventually receives stacking rewards. Voters have only 1 vote, no way to change.

setzeus:
Just double-checking here the logic here. It's okay that a registered stacker extends for the (eventually) losing POX address because the delegate aggregating committing overrides that POX? 

Reviewing POX-3 logic but want to make sure I understand.

friedger:
Once voted for a loosing wallet, there is no way back. The commit only ensure that the stx are earmarked for earning rewards, it does not change the locking state. 

All stackers should agree on one wallet and then do the vote. 

Is the new peg wallet deterministic in some way?

setzeus:
Unsure how FROST is used to generate the peg-wallet candidates. 

Reviewing (delegate-stack-extend ...) I'm wondering if there's any reason we can't just mass delegate-stack-extend the moment a successful wallet (> 70% consensus) is voted upon?

As long as we're recording that a stacker voted, once a vote is successful in consensus we can (delegate-stack-extend ...) to the winner for everyone regardless of who they voted for - any reason this approach wouldn't work?

friedger:
Yes, that can work. Good idea!

friedger:
We have to see how many delegate call we can get into one txs resource wise

jo-tm:
will be dealt in #578 a function take care of this. Kudos to friedger.

friedger:
fixed by #578

Issue #547, opened 2023-07-13
[stacking-pool] penalty-unhandled-request does not lock stackers to burn address


friedger:
**Describe the bug**
The function `penalty-unhandled-request` does not lock any stx, it only calls stack aggregation commit for the burn address.

**To Reproduce**
Inspect `penalize-helper`
https://github.com/Trust-Machines/stacks-sbtc/blob/bb5ed58befb5305f6ea4688e2e6efde8f328c0f7/sbtc-mini/contracts/sbtc-stacking-pool.clar#L596

**Expected behavior**
stx are locked for burn address.
delegate-stx-extend should be called with the burn address.

Instead of burning btc we should not call stack-aggregation-commit for the burn address, then the burn address won't get any rewards and the community of stackers will benefit more.

jo-tm:
@setzeus bug report from friedger, if you need clarification we can discuss next Tuesday.

setzeus:
@friedger made a commit here to address : https://github.com/Trust-Machines/stacks-sbtc/tree/sbtc-mini-stacking-pool-quick-fixes

Updated (penalize-helper ...) by removing the aggregate-commit & by adding a penalize-mass-delegate-stack-extend function to fold through & instead delegate-stack-extend for every stacker. 

Is this the solution you were envisioning?

friedger:
Yes, that could work as long as stackers don't revoke.

In pox-4, we feed to remove that revoke function.

You need to check the results of the extend function @setzeus

jo-tm:
also included in #578

friedger:
fixed by https://github.com/Trust-Machines/stacks-sbtc/pull/578

PR #578, opened 2023-07-17
sBTC-mini Stacking Pool Quick Fixes

setzeus:
Minor/Medium changes to address [562](https://github.com/Trust-Machines/stacks-sbtc/issues/562), [547](https://github.com/Trust-Machines/stacks-sbtc/issues/547), & [548](https://github.com/Trust-Machines/stacks-sbtc/issues/548). 

Main change is the addition of a private helper (mass-delegate-stack-extend ...) that extends stacking to either the burn address (in the form of penalize-helper) or to the newly voted-upon sBTC-wallet (in the form of a consensus vote).

Since the private function runs in different places, the response handling is inconsistent & could probably be "prettier" if any review has a suggestion for line 729.


setzeus:
@FriendsFerdinand neither of these updates changes the minimal unit tests found in sbtc-stacking-pool_test.clar. We're still working on how to best POX-3 & other contracts testing 2.0 in another branch.

Fixes here are mainly to address the minor/medium issues flagged in linked issues. Minor one from Mike was removing a dummy/test function I left in a WIP. 

The two medium issues flagged by Friedger had to do with me missing "delegate-stack-extend" calls in both voting & penalizing. This PR addresses both of those issues by using a private (mass-delegate-stack-extend ...) found in line 738; this is called from either (vote-for-threshold-wallet-candidate ...) once the next peg-wallet has found voting consensus, or from one of the four (penalize-helper ...) functions once a penalty is successfully submitted.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -729,14 +726,23 @@
             rewards-disbursed: bool
         }
     ))
-    (match (as-contract (contract-call? .pox-3 stack-aggregation-commit-indexed pox-burn-address next-cycle))
+    (try! (match (fold mass-delegate-stack-extend (get stackers penalized-pool) (ok {stacker: tx-sender, unlock-burn-height: u0, pox-addr: pox-burn-address}))
```
friedger, line 729: Why not just return the result of the fold call?

jo-tm:
We are following the Diagram in the Main Doc, STACKING:
image

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -540,19 +522,33 @@
...
-                            false
+
+                    ;; 70% consensus reached, ready to set next cycle threshold-wallet
+                    ;; Extend delegate all signers in pool & attempt to aggregate-commit-index
+                    (begin 
+                        
+                        ;; Assert that extend-and-commit-bool is true (both delegate-stack-extend & aggregate-commit-indexed succeeded)
+                        (asserts! (match (fold mass-delegate-stack-extend next-pool-stackers (ok {stacker: tx-sender, unlock-burn-height: u0, pox-addr: pox-addr}))
+                            passed-result
+                                (match (as-contract (contract-call? .pox-3 stack-aggregation-commit-indexed pox-addr next-cycle))
```
friedger, line 533: `stack-aggregation-commit-indexed` should be called the first time the 70% condition passed, thereafter `stack-aggregation-commit-increase` needs to be called with the index from the first call.

See https://github.com/friedger/clarity-stacking-pools/blob/main/contracts/pox-pool-self-service.clar#L175

jo-tm:
SBTC MINTING (PEG-IN):
image

friedger:
To be able to handle the penalty we must to not lock the stackers during the voting window.
Instead, we should call the extend with either burn address or peg-wallet during the penalty window.

The penalty window must start before the pox prepare phase. Maybe 200 blocks before the start of the cycle.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Thinking through this I don't believe we need `stack-aggregation-commit-increase`?

1. The pool of stackers registered in the previous cycle
2. Whether they vote for the winning wallet, they vote for the losing wallet, or they _abstain_ from the vote, they should still be part of the extend-delegate + aggregate commit with the winning wallet
3. Once the winning wallet reaches consensus (> 70% stacking weight), we don't care for additional votes
4. Only registered signers can vote & that list of registered signers is fixed (aka no one can register & vote in the same cycle)

Therefore once we mass-delegate-extend & aggregate-commit we should never need to aggregate-commit-increase? At least within the voting logic*

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -729,14 +726,23 @@
...
+            (ok (ok u1))
         err-result
-            false
+            err-mass-delegate-stack-extend
+    ))
+)
+
+;; mass delegate stack extender helper for either penalizing or concluded vote
+(define-private (mass-delegate-stack-extend (stacker principal) (pox-return (response {stacker: principal, unlock-burn-height: uint, pox-addr: { version: (buff 1), hashbytes: (buff 32) }} uint))) 
+    (let
```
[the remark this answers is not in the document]
setzeus: Didn't get picked up but took care of this [in latest commit](https://github.com/Trust-Machines/stacks-sbtc/pull/578/commits/0df2d20e75f3e8c9d350ec981b4ae9ad4f9f7a9c) with a simple asserts: 

;; Assert previous pox-return was (ok ...)
`(asserts! (is-ok pox-return) err-mass-delegate-stack-extend)`

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Severely overcomplicated myself, appreciate the suggestion.

sbtc-mini/contracts/sbtc-stacking-pool.clar
friedger: Ok, make sense. If voting is not required for signers.

friedger:

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -63,6 +63,8 @@
 (define-constant err-not-protocol-caller (err u35))
 (define-constant err-threshold-percent-out-of-range (err u36))
 (define-constant err-threshold-to-scriptpubkey (err u37))
+(define-constant err-mass-delegate-stack-extend (err u38))
+(define-constant err-wallet-consensus-reached-execution (err u39))
```
MarvinJanssen, line 67: Just an FYI for everyone. We need to formally lay down all protocol error codes. I will bring it up the next meeting.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -497,11 +477,12 @@
             next-pool-signer 
             {vote: (some pox-addr)}
         ))
+        
 
         ;; Check whether map-entry for candidate-wallet already exists
         (if (is-none current-candidate-status)
```
MarvinJanssen, line 483: This entire function is pretty big with a lot of nesting. Is it possible to split it up into a few functions for readability?

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -540,19 +522,33 @@
...
+                            err-result
+                                false
+                        ) err-wallet-consensus-reached-execution)
+
+                        true
                     )
+
                     ;; 70% consensus not reached, continue
-                    false
+                    true
```
MarvinJanssen, line 551: This nested structure is pretty hard to follow. (Especially the many else/match err branches.) Can it be refactored to a series of `asserts!`?

MarvinJanssen (changes requested):
Looks good, comments are mostly readability questions.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Refactored with series of asserts, more in line with what you were thinking?

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Yeap, updated whenever you get a chance to approve.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -497,68 +478,66 @@
...
-            )
 
-            ;; Candidate exists, update relevant maps: votes-per-cycle, signer, & check whether this wallet is over threshold
-            (let
+            )
+            ;; Existing candidate path
+            (let 
                 (
-                    (unwrapped-candidate (unwrap! current-candidate-status err-unwrapping-candidate))
+                    (unwrapped-candidate (unwrap-panic current-candidate-status))
```
friedger, line 504: current-candiate-status is-none, isn't it?
As checked here: https://github.com/Trust-Machines/stacks-sbtc/pull/578/files#diff-9e54e6dfb041bc4c76781bc43adfdf0ce7f1b980d41f9207780022f52d558cf8R487

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -497,68 +478,66 @@
...
-                            ))
-                        err-result
-                            false
-                    )
-                    ;; 70% consensus not reached, continue
-                    false
-                )
-
+                                    next-pool
+                                    {last-aggregation: (some block-height), reward-index: (some (unwrap-panic (as-contract (contract-call? .pox-3 stack-aggregation-commit-indexed pox-addr next-cycle))))}
```
friedger, line 530: I would put the result of the contract call into a separate line/variable.
unwrap-panic seems wrong as stack-aggregation-commit can return an error.

friedger (commented):
Use of unwrap-panic is not clear to me.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Negative, I think? The (let ...) runs when one of the (and ...) conditions fail. I check (is-none current-candidate=status) is true as the first condition.

If that returns as false, aka current-candidate-status *does exist* then that triggers the (let ...) branch. Since we know current-candidate-status is a (some ...) we can safely unwrap-panic.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -497,68 +478,66 @@
...
+                        aggregate-commit-index: none,
+                        votes-in-ustx: (get amount next-pool-signer),
+                        num-signer: u1,
                 })
-                
                 ;; Update pool map by appending wallet-candidate to list of candidates
                 (map-set pool next-cycle (merge 
                     next-pool
-                    {threshold-wallet-candidates: (unwrap! (as-max-len? (append (get threshold-wallet-candidates next-pool) pox-addr) u100) err-too-many-candidates)}
+                    {threshold-wallet-candidates: (unwrap-panic (as-max-len? (append (get threshold-wallet-candidates next-pool) pox-addr) u100))}
```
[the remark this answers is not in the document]
setzeus: Copy. Changing back. I think I'm now swinging the other way on using unwrap-panic instead of overusing (unwrap! ...). 

Any chance you could summarize the conditions on when you find it appropriate to specifically unwrap-panic?

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Changed back to a match, will only use panic sparingly / as we define.

sbtc-mini/contracts/sbtc-stacking-pool.clar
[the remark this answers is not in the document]
friedger: I would use unwrap-panic only when I am sure that panic won't happen.

sbtc-mini/contracts/sbtc-stacking-pool.clar
friedger: OK, the git diff was not clear to me.

sbtc-mini/contracts/sbtc-stacking-pool.clar
friedger: Added an issue: https://github.com/Trust-Machines/stacks-sbtc/issues/590

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -497,68 +479,71 @@
...
+                            )
+                        )
+                    (ok u1))
+                    
+                    (ok u2)
             )
         )
 
-        (ok true)
+        (ok u3)
```
friedger, line 546: Codes should be constants with some meaningful names.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -729,14 +715,22 @@
...
-        err-result
-            false
+    (fold mass-delegate-stack-extend (get stackers penalized-pool) (ok {stacker: tx-sender, unlock-burn-height: u0, pox-addr: pox-burn-address}))
+)
+
+;; mass delegate stack extender helper for either penalizing or concluded vote
+(define-private (mass-delegate-stack-extend (stacker principal) (pox-return (response {stacker: principal, unlock-burn-height: uint, pox-addr: { version: (buff 1), hashbytes: (buff 32) }} uint))) 
+    (let
+        (
+            (extend-delegate-response (unwrap! (as-contract (contract-call? .pox-3 delegate-stack-extend stacker pox-burn-address u1)) err-mass-delegate-stack-extend))
```
friedger, line 725: Idea for optimization: use `as-contract` around `(fold mass-delegate-stack-extend` To be verified whether it makes a difference.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -497,68 +479,71 @@
...
+                            (>= (/ (* new-candidate-votes u1000) next-pool-total-stacked) (var-get threshold-consensus))
+                            
+                            (match (as-contract (contract-call? .pox-3 stack-aggregation-commit-indexed pox-addr next-cycle))
+                                ok-index
+                                    (map-set pool next-cycle (merge 
+                                        next-pool
+                                        {last-aggregation: (some block-height), reward-index: (some ok-index)}
+                                    ))
+                                err-result
+                                    ;; Chore - anything additional in this branch?
```
friedger, line 536: We could emit a print event

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -497,68 +479,71 @@
...
-        (if (is-none current-candidate-status)
-
-            ;; Candidate doesn't exist, update relevant maps: votes-per-cycle & pool
-            (begin 
-                ;; Update votes-per-cycle map with first vote for this wallet-candidate
-                (map-set votes-per-cycle {cycle: next-cycle, wallet-candidate: pox-addr} {
-                    aggregate-commit-index: none,
-                    votes-in-ustx: (get amount next-pool-signer),
-                    num-signer: u1,
+        (asserts!
```
friedger, line 484: I find this control flow hard to understand. Suggestion:

```
(if (is-none current-candidate-status) (add-wallet-candiate ....) (update-wallet-candidate...))
(maybe-aggregate-commit...)
```

Can a whale trigger the aggregate commit? If not

```
(if (is-none current-candidate-status) 
   ;; handle new candidate
   (add-wallet-candiate ....) 
   ;; handle existing candidate
   (begin
     (update-wallet-candidate...)
     (maybe-aggregate-commit...)))
```

friedger (changes requested):
Needs rebase.

I have an issue with the control flow of signer-pre-register

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Great idea, doing so.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Copy. Err responses are clear to me how to identify with uint, but how/what would you use to identify an (ok ...) response? 

Same thing as error? Just a number saved as more meaningful string constant?

friedger: Yes, like `(ok result-committed)`

friedger: Or better `ok-committed`

MarvinJanssen: Given our design pattern for error constants, I think the latter makes the most sense.

```clarity
(define-constant ok-committed (ok u3))
```

sbtc-mini/contracts/sbtc-stacking-pool.clar
MarvinJanssen: It should make a difference.

sbtc-mini/contracts/sbtc-stacking-pool.clar
MarvinJanssen: Good idea, we do not have a comprehensive spec for custom events yet. (Something I will bring up later.)

sbtc-mini/contracts/sbtc-stacking-pool.clar
MarvinJanssen: Both if branches can be abstracted into two private functions.

setzeus: Removed the (if ...) that was previously there in exchange for that series of asserts! Should I change back?

Re: Hoisting Out Private Funcs
Great discussion point here. My personal rule-of-thumb is it's harder to hop around private-functions than it is to read a single well-commented / well-labeled function. 

If I need a series of functions for more than one place, of course I'll hoist out into a private function to no repeat, otherwise though, if it's only callable by one other function I'll leave it as-is. 

Thoughts here? Good opportunity to capture & document in the new docs Marvin introduced.

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Wasn't picked up but updated with the (as-contract ...) wrapped around the (fold ...).

sbtc-mini/contracts/sbtc-stacking-pool.clar
friedger: I like to read in particular public functions without going into unnecessary details. I think that is also true for the user of a contract.
In a second go I would verify whether the private functions indeed implement what the name suggests.

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -523,25 +523,21 @@
...
-                    (ok u1))
+                    ok-vote-existing-candidate-lost)
                     
-                    (ok u2)
+                    ok-vote-existing-candidate-won
             )
         )
 
-        (ok u3)
+        ok-voted
```
MarvinJanssen: These nested `asserts!` are still a little tricky to follow. It is not immediately clear what these short-circuit branches belong to.

sbtc-mini/contracts/sbtc-stacking-pool.clar
friedger: The `as-contract` should be removed here. @setzeus

sbtc-mini/contracts/sbtc-stacking-pool.clar
friedger: Let's move this discussion to the document or somewhere else and let's move this PR forward.

I suggest to merge this PR and open a new issue for better readability.

friedger (changes requested):
The last thing I would like to see is the removal of `as-contract` in `mass-delegate-stack-extend` and tests are failing.

sbtc-mini/contracts/sbtc-stacking-pool.clar
friedger: We can improve readability in a new PR

sbtc-mini/contracts/sbtc-stacking-pool.clar
friedger: events would be out of scope for this PR

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Not picked up but now removed, all conversations should be resolved / ready.

friedger (approved):
Let's merge this and optimize in a separate PR

friedger:
@MarvinJanssen Do you have time to review?

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -495,68 +482,81 @@
...
+                        aggregate-commit-index: none,
+                        votes-in-ustx: (get amount next-pool-signer),
+                        num-signer: u1,
                 })
-                
                 ;; Update pool map by appending wallet-candidate to list of candidates
                 (map-set pool next-cycle (merge 
                     next-pool
-                    {threshold-wallet-candidates: (unwrap! (as-max-len? (append (get threshold-wallet-candidates next-pool) pox-addr) u100) err-too-many-candidates)}
+                    {threshold-wallet-candidates: (unwrap! (as-max-len? (append (get threshold-wallet-candidates next-pool) pox-addr) u100) err-candidates-overflow)}
```
FriendsFerdinand, line 501: There's nothing stopping a signer with a low stake in the protocol to fill the entire list with rubbish addresses.

setzeus: In line 477 I assert that signer hasn't yet voted. 

So I believe that at most one signer with a low stake would be able to submit a single rubbish address? Or am I misunderstanding?

Either way we could / should check against the address by version I believe (though agreed I'll open another issue for that so this can get approved merged).

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -495,68 +482,81 @@
...
+                                        (begin 
+                                            (print (/ (* new-candidate-votes u1000) next-pool-total-stacked))
+                                            false
+                                        )
+                                    )
+                                err-result
+                                    false
+                            )
+                        )
+                    ok-vote-existing-candidate-lost)
```
FriendsFerdinand, line 553: This is the outcome of the branch where where someone voted correctly, no?
Why are we adding the `-lost` part here? That seems to indicate that you can't vote anymore after someone has lost.

setzeus: Idea here is to keep branches returning parallel (ok ...) responses. 

Since a signer is voting for an existing candidate, that vote either pushes the existing candidate over the threshold-consensus, or it doesn't. 

If it does push existing candidate past consensus, we return (ok-vote-existing-candidate-won ...) - which is an ok vote for an existing candidate that won. 

If it doesn't push the existing candidate past consensus, we return (ok-vote-existing-candidate-lost ...) - which is an ok vote for an existing candidate that lost / didn't push the candidate past consensus.

Open to other naming conventions for these responses if you have any particular suggestion in mind.

FriendsFerdinand: @setzeus Ok, it should be called `ok-vote-existing-candidate`

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -727,14 +728,22 @@
...
+    (as-contract (fold mass-delegate-stack-extend (get stackers penalized-pool) (ok {stacker: tx-sender, unlock-burn-height: u0, pox-addr: pox-burn-address})))
+)
+
+;; mass delegate stack extender helper for either penalizing or concluded vote
+(define-private (mass-delegate-stack-extend (stacker principal) (pox-return (response {stacker: principal, unlock-burn-height: uint, pox-addr: { version: (buff 1), hashbytes: (buff 32) }} uint))) 
+    (let
+        (
+            (extend-delegate-response (unwrap! (contract-call? .pox-3 delegate-stack-extend stacker pox-burn-address u1) err-mass-delegate-stack-extend))
+            (extend-delegate-stacker (get stacker extend-delegate-response))
+            (extend-delegate-unlock-burn-height (get unlock-burn-height extend-delegate-response))
```
FriendsFerdinand, line 740: Why are we always setting the `pox-addr` to the `pox-burn-address` ? Don't we have to use the `pox-addr` in the parameters?

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -727,14 +728,22 @@
...
+        )
+
+        ;; Assert previous pox-return was (ok ...)
+        (asserts! (is-ok pox-return) err-mass-delegate-stack-extend)
```
FriendsFerdinand, line 744: We can `unwrap!` instead of doing 2 commands (`asserts!` and `is-ok`)

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Yes. Great catch here, forget to update accordingly. Now outdated in latest commit:

https://github.com/Trust-Machines/stacks-sbtc/pull/578/commits/47e25a3065373ce482534db5410485e808892f93

sbtc-mini/contracts/sbtc-stacking-pool.clar
setzeus: Copy, fixed.

FriendsFerdinand (approved):

MarvinJanssen:
Thanks @friedger

PR #579, opened 2023-07-17
Improve get-current-window

### LINKED ISSUE #558: [sbtc-stacking-pool] get-current-window throws execution cost error
**Call to get-current-window fails**

```
callContractReadOnly: http://devnet.stx.eco/v2/contracts/call-read/ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM/sbtc-stacking-pool/get-current-window
callContractReadOnly:  {
  okay: false,
  cause: 'Unchecked(CostBalanceExceeded(ExecutionCost { write_length: 0, write_count: 0, read_length: 109265, read_count: 10, runtime: 111659 }, ExecutionCost { write_length: 0, write_count: 0, read_length: 100000, read_count: 30, runtime: 1000000000 }))'
}
```

**To Reproduce**
This api call wraps a call to get-current-window as sbtcWindow.

friedger:
This PR
* copies the conversion of cycle-to-burn-height and back from pox-3 contract to sbtc-stacking-pool
* reduces the read-length to > 40k (from 200k+) which is below the standard read-only limit
* removes current-window let assignment because it will result in early exit
* removes unnecessary comparison in the assertions for the end of the corresponding windows
* fixes #558


sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -24,6 +24,8 @@
 ;; Dust limit placeholder for checking that pox-rewards were disbursed (in sats)
 (define-constant dust-limit u100)
 
+(define-constant pox-info (unwrap-panic (contract-call? .pox-3 get-pox-info)))
```
setzeus, line 27: Curious on why unwrap-panic instead of unwrap? Generally prefer to explicitly throw for bug-tracking.

setzeus (commented):
Confirmed checked out & all pertinent unit tests running. 

No changes requested but since we're likely using that unwrap-panic / pox-info constant in various places & need to refactor accordingly I'd like to better understand unwrap-panic vs. unwrap! with an explicit error.

sbtc-mini/contracts/sbtc-stacking-pool.clar
friedger: There are no reasons why get-pox-info would return an error. Therefore we can safely use unwrap-panic without loosing any information.

Furthermore, if we use unwrap!, the error would not appear anywhere where we could check I think. To be verified. 

With unwrap-panic the deployment tx would return (err none)

setzeus: Ah, yeap, okay. 

Erroneously thought it could return an error but just re-read (get-pox-info ...) in pox-3 & understand now. Thank you for clarifying, making an issue update the rest of the contract with this function wherever it makes sense.

setzeus (approved):
Discussed, pulled & confirmed that all new/previous unit tests (59) intact. Lgtm.

friedger:

PR #580, opened 2023-07-17
Add signer api configuration management

jferrant:


stacks-signer-api/src/db/mod.rs
```diff
@@ -21,34 +18,52 @@ pub enum Error {
...
+            FOREIGN KEY(signer_id) REFERENCES signers(signer_id) ON DELETE CASCADE
+        );
+        "#;
+
+// SQL schema for creating the `transactions` table
 const SQL_SCHEMA_TRANSACTIONS: &str = r#"
-        CREATE TABLE transactions (
-            txid TEXT PRIMARY KEY,
+        CREATE TABLE IF NOT EXISTS transactions (
+            txid TEXT NOT NULL PRIMARY KEY,
```
netrome, line 66: Shouldn't this be deleted in favor of the migrations?

stacks-signer-api/src/routes/signers.rs
```diff
@@ -157,13 +193,92 @@ pub async fn get_signers(
...
+    signer_id: i64,
+    authorization_value: String,
+    pool: SqlitePool,
+) -> Result<Box<dyn Reply>, Infallible> {
+    if let Ok(signer) = db::signers::get_signer_by_id(&pool, signer_id).await {
+        let public_key = signer.config.public_key;
+        let message =
+            Message::from_hashed_data::<sha256::Hash>(signer.signer_id.to_string().as_bytes());
+        if let Ok(sig) = Signature::from_str(authorization_value.as_str()) {
+            match sig.verify(&message, &public_key) {
```
netrome, line 249: Is this procedure documented anywhere? From what I see we're just using the signature of the signer ID as authentication mechanism, which does not seem like a particularly good solution. This is at least vulnerable to replay attacks.

I'm also curious to know the threat model we're protecting against. If this API is running locally on the signers own infra, why do we need authentication and authorization in the first place?

netrome (commented):
The vast majority of the code looks good. My main concern here is how authentication is handled and why it is needed. I'd like to gain some clarity on that before I feel comfortable approving this.

stacks-signer-api/src/routes/signers.rs
jferrant: Yes it def is, I haven't added a timestamp to it yet :) There is a ticket to add proper auth to the code but I added this because it was quick and easy.

Auth is required because no one should be able to access the config of a signer accept the owner of said signer.

netrome: Alright, so this is a quick and dirty first stab at auth which will be improved in the subsequent ticket. Makes sense. Thank you for the clarification!

netrome (approved):

stacks-signer-api/src/db/mod.rs
jferrant: Good point. Done :)

netrome: Perfect, thank you! :pray:

stacks-signer-api/src/routes/signers.rs
jferrant: I will add a TODO to make it extra explicit that it is meant to be placeholder code!

netrome: Bonus points if you link the issue in the TODO like `//TODO(#XXX): ...`

jferrant: Haha I will do so now XD

stjepangolemac (approved):

Issue #549, opened 2023-07-13
[stacking-pool] Add test for stacking


friedger:
**Is your feature request related to a problem? Please describe.**
clarinet unit tests does not support stacking properly, stx-account does not handle locked amount

**Describe the solution you'd like**
Implement mock behaviour in .pox-3 that allows tests to set locked amount manually.

**Describe alternatives you've considered**
Use integration tests on devnet.

**Additional context**
https://github.com/Trust-Machines/stacks-sbtc/pull/544#issuecomment-1633579268

jo-tm:
will be solved in a branch from @setzeus , no PR yet.

friedger:
fixed by #613

Issue #582, opened 2023-07-18
Fix CI time bomb caused by https://github.com/rust-bitcoin/rust-miniscript/issues/549


netrome:
Our CI is failing since it updated to rustc 1.71. The cause is the upstream `miniscript` library which doesn't build with 1.71.

Example failure:
https://github.com/Trust-Machines/stacks-sbtc/actions/runs/5574993273/jobs/10187283929?pr=576

How to reproduce locally:
```
cargo build -p sbtc-cli  --release
```

### Definition of done
Our CI is capable of building our code.

netrome:
Given how many time bombs we have encountered in our CI lately, I am strongly suggesting revisiting our conclusions in this discussion. https://github.com/orgs/Trust-Machines/discussions/309

Ping @sergey-shandar

PR #585, opened 2023-07-18
Add mock for stx-account in pox-3

friedger:
This PR
* adds a setter and getter functions for the unlocked, locked amount and unlock-height for any users in pox-3
* replaces `stx-account` with `get-stx-account` in pox-3
* see #549

### DIFF
--- a/sbtc-mini/contracts/pox-3.clar
+++ b/sbtc-mini/contracts/pox-3.clar
@@ -224,6 +224,19 @@
     { amount: uint }
 )
 
+;; MOCK
+;; Allow to set stx-account details for any user
+;; These values are used for PoX only
+(define-map stx-account-details principal {unlocked: uint, locked: uint, unlock-height: uint})
+
+(define-read-only (get-stx-account (user principal))
+    (default-to (stx-account user) (map-get? stx-account-details user)))
+
+(define-public (set-stx-account (user principal) (details {unlocked: uint, locked: uint, unlock-height: uint}))
+    (if (map-set stx-account-details user details)
+        (ok true) (err u9999))) ;; define manually the error type
+;; MOCK END
+
 ;; Getter for stacking-rejectors
 (define-read-only (get-pox-rejection (stacker principal) (reward-cycle uint))
     (map-get? stacking-rejectors { stacker: stacker, reward-cycle: reward-cycle }))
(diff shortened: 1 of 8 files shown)


friedger:
Tests will be added when the new test generation has landed.

friedger:

sbtc-mini/ext/generate-tests.ts
```diff
@@ -93,9 +93,14 @@ function generatePrepareTx(contractPrincipal: string, annotations: FunctionAnnot
...
 }
 
+function generatePox3AllowContractCaller(annotations: FunctionAnnotations) {
+	return `Tx.contractCall(deployer.address + '.pox-3', 'allow-contract-caller', [ "'" + deployer.address + '${annotations['poxadmin']}', 'none'], callerAddress)`;
+}
+
 function generateNormalMineBlock(contractPrincipal: string, testFunction: string, annotations: FunctionAnnotations) {
 	return `let block = chain.mineBlock([
 		${annotations['prepare'] ? `${generatePrepareTx(contractPrincipal, annotations)},` : ''}
+		${annotations['poxadmin'] ? `${generatePox3AllowContractCaller(annotations)},` : ''}
```
MarvinJanssen, line 103: Nice solution but what I do not like is that it makes the generation script more opinionated. How would you feel about an annotation that calls a custom TS function? (Just how `bootstrap.ts` is separated now.)

sbtc-mini/contracts/pox-3.clar
```diff
@@ -224,6 +224,19 @@
...
 
+;; MOCK
+;; Allow to set stx-account details for any user
+;; These values are used for PoX only
+(define-map stx-account-details principal {unlocked: uint, locked: uint, unlock-height: uint})
+
+(define-read-only (get-stx-account (user principal))
+    (default-to (stx-account user) (map-get? stx-account-details user)))
+
+(define-public (set-stx-account (user principal) (details {unlocked: uint, locked: uint, unlock-height: uint}))
```
MarvinJanssen, line 235: Can we prefix these functions names with `mock-` or `debug-`? I prefer the latter I think. It is something like that in `clarity-bitcoin` too.

MarvinJanssen (commented):
Did a quick pass. Will do a full review when you are unblocked and can write all tests.

sbtc-mini/ext/generate-tests.ts
friedger: This should not be part of the PR. New PR sent with changes only to pox-3.

sbtc-mini/contracts/pox-3.clar
friedger: adding `mock-`

friedger:
Replaced by #586

PR #586, opened 2023-07-19
Add mock function for stx-account in pox-3

friedger:
This PR
* adds a setter and getter functions for the unlocked, locked amount and unlock-height for any users in pox-3
* replaces `stx-account` with `get-stx-account` in pox-3
* see #549


sbtc-mini/contracts/pox-3.clar
```diff
@@ -224,6 +224,20 @@
...
+;; Allow to set stx-account details for any user
+;; These values are used for PoX only
+(define-map mock-stx-account-details principal {unlocked: uint, locked: uint, unlock-height: uint})
+
+(define-read-only (get-stx-account (user principal))
+    (default-to (stx-account user) (map-get? mock-stx-account-details user)))
+
+(define-public (mock-set-stx-account (user principal) (details {unlocked: uint, locked: uint, unlock-height: uint}))
+    (if (map-set mock-stx-account-details user details)
+        (ok true) (err u9999))) ;; define manually the error type
```
netrome, line 238: Nit, I'm not a clarity expert but I assume it would be better to define a constant for this error type like the other error types.

netrome (approved):
LGTM, but also not my area of expertise.

sbtc-mini/contracts/pox-3.clar
friedger: My thinking was that this error does not happen anywhere and therefore, the additional constant would only add unnecessary cognitive load

PR #589, opened 2023-07-19
Fix #588: sBTC alpha coordinator fulfills withdrawals to itself

### LINKED ISSUE #588: sBTC alpha coordinator fulfills withdrawals to itself
The alpha coordinator is capable of fulfilling withdrawal requests beautifully except for a teeny tiny detail; Instead of sending the withdrawal amount to the recipient denoted in the withdrawal request - it sends it back to the dkg wallet.

So instead of doing
```
        let withdrawal_output = bitcoin::TxOut {
            value: op.amount,
            script_pubkey: script_pubkey.clone(), // This is where we're paying to ourselves.
        };
```
we could simply do

```
    let withdrawal_output = op.recipient.to_bitcoin_tx_out(op.amount)
```

### Definition of done
The alpha coordinator fulfills withdrawals by sending the requested amount to the intended recipient instead of itself.

igorsyl:


xoloki:
Is there a unit or integration test that checks to see that this works, and the output actually goes to the correct place?  Presumably not, since we didn't realize this was broken.

igorsyl:
> Is there a unit or integration test that checks to see that this works, and the output actually goes to the correct place? Presumably not, since we didn't realize this was broken.

I don't think so. I'll work on Improving test coverage as my primary focus.

stacks-coordinator/src/bitcoin_wallet.rs
```diff
@@ -107,11 +107,12 @@ impl BitcoinWalletTrait for BitcoinWallet {
...
 
-        let withdrawal_output = bitcoin::TxOut {
+        let withdrawal_output_stacks = op.recipient.to_bitcoin_tx_out(op.amount);
+        let withdrawal_output_bitcoin = bitcoin::TxOut {
             value: op.amount,
-            script_pubkey: script_pubkey.clone(),
+            script_pubkey: Script::from(withdrawal_output_stacks.script_pubkey.into_bytes()),
         };
-        tx.output.push(withdrawal_output);
+        tx.output.push(withdrawal_output_bitcoin);
```
netrome, line 115: Nit: `.to_bitcoin_tx_out` already returns a TxOut so this can be simplified as
```
tx.output.push(op.recipient.to_bitcoin_tx_out(op.amount));
```

netrome (approved):
The solution seems correct but could be simplified. Also it would be great to get this tested before merging.

stacks-coordinator/src/bitcoin_wallet.rs
igorsyl: These are two different TxOut traits.

netrome: Yeah right, because the Stacks blockchain doesn't use the upstream Bitcoin lib but instead copied it... Ugh :sweat_smile:

netrome:
@igorsyl any updates on this? Are you still working on testing? Do you need any help or pointers?

netrome:
This is deployed on the alpha VM and not needed for further development. Closing.

Issue #519, opened 2023-07-06
[BLOCKER] Clarity Native Testing Better Annotations


jo-tm:

jo-tm:
@setzeus pls write 1 line or 2 on this issue description.

jo-tm:
@setzeus : working on test annotations for mining empty blocks to push the chain ahead an being in another window, blocking other tasks.

jo-tm:
work in progress in test generation branch.

friedger:
Definition of done:
clarity unit tests can be written in clarity like this:
```
(define-public (test-pox)
   ;; @external
   ;; @caller wallet_1
   (try! (contract-call? .pox-3 allow-contract-caller 'ST1234 none))

  ;; @mine-empty 5

  (contract-call? .sbtc-stacking-pool do-something)
) 
```
resulting in 
```
let block = chain.mineBlock([
		Tx.contractCall('ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.pox-3', 'allow-contract-caller', [...], callerAddress)]);
chain.mineEmptyBlock(5);
block = chain.mineBlock([
		Tx.contractCall('ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.sbtc-stacking-pool_flow_test', 'test-pox_part2', [], callerAddress)]);
```

@MarvinJanssen @setzeus Is that the correct definition of done?

friedger:
Or something like
```
(define-public (test-pox)
   ;; @new-block
   ;; @caller wallet_1
   (try! (contract-call? .pox-3 allow-contract-caller 'ST1234 none))

  ;; @mine-empty 5

  ;; @new-block
  (do-and-assert-something)

  ;; @mine-empty 5

  (do-and-assert-something)
) 
```

resulting in

```
let block = chain.mineBlock([
		Tx.contractCall('ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.pox-3', 'allow-contract-caller', [...], callerAddress)]);
block.receipts[0].expectOk();

chain.mineEmptyBlock(5);
block = chain.mineBlock([
		Tx.contractCall('ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.sbtc-stacking-pool_flow_test', 'do-and-assert-something', [], callerAddress)]);

chain.mineEmptyBlock(5);
[...]

PR #591, opened 2023-07-20
Add simple test for sign-pre-register

friedger:
This PR
* adds a simple test for sign-pre-register

see #549 

More tests when #519 solved.


MarvinJanssen:
Is this good for review? (Draft.)

LGTM!

setzeus (approved):
LTGM - the only thing I'd like to clarity/document for near-future work is what exactly goes into sbtc-stacking-pool_test vs. sbtc-stacking-pool_flow_test?

I'm assuming by "flow" we mean anything that requires a sequence of functions + chain advancement? 

Or how do you see this defined for future tests?

friedger:
@setzeus Yes, flow should be used for tests that mine tx in different blocks and make calls to other contracts. I renamed the test to `_signer_test`.

FriendsFerdinand (approved):

Issue #593, opened 2023-07-20
Ignore Github Workflow when Only Documentation Changes


AshtonStephens:
We've discussed that our CI is a barrier of entry to making minor alterations to the repo, even when only propose a that we prevent the github actions from running automatically when the related change includes only `md` files.

We should do this by adding the following line to our github action yml files that launch automatically.
```
on:
    {trigger}:
    ...
    paths-ignore: ['**.md']
```

PR #594, opened 2023-07-20
Ignore Github workflow when only MD files are altered. (#593)

AshtonStephens:
We've discussed that our CI is a barrier of entry to making minor alterations to the repo, even when only propose a that we prevent the github actions from running automatically when the related change includes only md files.

We should do this by adding the following line to our github action yml files that launch automatically.
```
on:
    {trigger}:
    ...
    paths-ignore: ['**.md']
```


AshtonStephens:
This should be linked to https://github.com/Trust-Machines/stacks-sbtc/issues/593.

friedger (approved):
:+1:

jo-tm (approved):
thanks!

PR #595, opened 2023-07-21
[Main] (Sponsored) withdrawal requests via Stacks side

MarvinJanssen:
This is a WIP / concept to explain my idea to be able to request peg-outs on the Stacks side. It requires sending a Stacks transaction, which contravenes the requirement of being able to do all sBTC actions on the Bitcoin side. However, by adding a simple sponsor mechanism, an sBTC holder that does not own any STX can still sign a Stacks transaction and pay transaction fees in sBTC. The basic mechanism is as follows:

1. sBTC holder creates and signs a sponsored Stacks contract call to `request-peg-out-sponsored`. The contract call specifies the amount, sender, and destination, as well as a fee in sBTC sats up for grabs to whoever sponsors the transaction.
2. The sBTC holder passes this transaction to a sponsor or sponsor pool of some kind, where sponsors can see it. (Out of scope for this PR.)
3. A sponsor can decide to sponsor the transaction if they find the fee acceptable. (Which is an argument of the contract call and thus set by the sBTC holder.)
4. The sponsor sponsors the transaction, the peg-out request is booked and the sponsor receives the sBTC fee payment.

I think it is a good alternative to starting the process on the Bitcoin chain, although we should still support that too. Starting on the Stacks side is most likely cheaper and quicker as well, as the sBTC holder does not have to fund a P2TR address with fees (which also costs fees) and wait for the signers to consume that output. With the proper UI/UX the process can be entirely transparent, with the user only meaningfully dealing with Bitcoin.


MarvinJanssen:
Tagging @netrome for visibility. @friedger do you join the WG syncs? If so, could you bring this concept up? It is too late for me to join (1am Saturday).

radicleart:
Is the sponsor ship required or are there  two options in 2)? They wait for sponsor or they send with 0 sponsor fee and pay the stx gas fee instead ?

MarvinJanssen:
Yes, there are two functions. You can either call it yourself directly and pay STX fees or generate a sponsored transaction to the other function that would cause you to pay an sBTC fee to the sponsor.

friedger:
@MarvinJanssen discussion is here https://github.com/stacks-network/sbtc-docs/issues/27

FriendsFerdinand:
Well, I think it makes sense now to include this option. Just need to resolve conflicts.

friedger:
Before merge, please add some unit tests.

MarvinJanssen:
This is currently being worked on. Branch will be updated as discussed.

MarvinJanssen:
@friedger can you take an early look? Should be good for end to end test. Some basic unit tests but we will need flow tests. We can decide if we want to merge soon and add more tests or add them on this branch. I prefer to merge since the PR is a month old.

sbtc-mini/contracts/sbtc-btc-tx-helper.clar
```diff
@@ -23,18 +23,18 @@
...
+		;; 	tx
+		;; 	header
+		;; 	tx-index
+		;; 	tree-depth
+		;; 	wproof
+		;; 	witness-merkle-root
+		;; 	witness-reserved-data
+		;; 	ctx
+		;; 	cproof
+		;; ))
```
MarvinJanssen, line 37: Note that the PR is WIP, still need to generate headers for the withdrawal unit tests and reenable the above.

sbtc-mini/tests/sbtc-stacking-pool_allow_flow_test.clar
```diff
@@ -66,23 +66,24 @@
...
+;;         (unwrap! (contract-call? .sbtc-stacking-pool allow-contract-caller 'ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.sbtc-stacking-pool_allow_flow_test (some u7)) (err u113))
+;;         ;; @mine-blocks-before 5
+;; 		(try! (check-sign-pre-register-disallowed))
+;;         ;; @continue
+;;         (unwrap! (contract-call? .sbtc-stacking-pool allow-contract-caller 'ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.sbtc-stacking-pool_allow_flow_test (some u9)) (err u114))
+;;         ;; @mine-blocks-before 1
+;; 		(try! (check-sign-pre-register-allowed))
+;;         (ok true)
+;;     )
+;; )
```
MarvinJanssen, line 86: This test was merged in but was failing for me. Disabled temporarily.

sbtc-mini/tests/sbtc-stacking-pool_test.clar
```diff
@@ -83,22 +83,22 @@
...
+;;             (start-transfer-window (- next-cycle-burn-height (+ normal-transfer-period-len normal-penalty-period-len)))
+;;             (start-penalty-window (- next-cycle-burn-height normal-penalty-period-len))
+;; )
+;; 		(asserts! (is-eq start-voting-window u3700) (err start-voting-window))
+;; 		(asserts! (is-eq start-transfer-window u4000) (err start-transfer-window))
+;; 		(asserts! (is-eq next-cycle-burn-height u4200) (err next-cycle-burn-height))
+;; 		(asserts! (is-eq burn-block-height u3701) (err burn-block-height))
+;; 		(ok true)
+;; 	)
+;; )
```
MarvinJanssen, line 101: This test was merged in but was failing for me. Disabled temporarily.

sbtc-mini/scripts/test.sh
```diff
@@ -3,5 +3,6 @@ mkdir -p .test
 mkdir -p .coverage
 rm -r .test/*
 clarinet run --allow-write --allow-read ext/generate-tests.ts
-clarinet run --allow-write --allow-read ext/generate-flow-tests.ts
+#TODO: flow test script never completes, makes Clarinet hang
+#clarinet run --allow-write --allow-read ext/generate-flow-tests.ts
```
MarvinJanssen, line 7: @friedger the flow test script hangs every now and then for me, so then the process never completes. Have you seen the issue before?

friedger: @MarvinJanssen Yes, if there is an error in the code.

sbtc-mini/contracts/sbtc-withdrawal-request-reclaim.clar
```diff
@@ -0,0 +1,17 @@
...
+(define-constant err-withdrawal-not-epxired (err u5300)) ;; The withdrawal request has not yet hit the expiry burn block height.
+
+;; Unlocks the sBTC tokens after expiry
+(define-public (reclaim-locked-tokens (withdrawal-request-id uint))
+	(let (
+		;; Get the pending withdrawal and mark it as settled.
+		;; The call will fail if the request is no longer pending.
+		(withdrawal-request (try! (contract-call? .sbtc-registry get-and-settle-pending-withdrawal-request withdrawal-request-id withdrawal-state-reclaimed)))
+		)
+		;; Check if the withdrawal request has expired (pending check is done above).
```
friedger, line 12: Why is this check not in the sbtc-registry?

sbtc-mini/contracts/sbtc-withdrawal-request-stx.clar
```diff
@@ -0,0 +1,68 @@
...
+(define-private (is-allowed-operator-and-deduct (sender principal) (operator principal) (amount uint))
+	(begin
+		(asserts! (not (is-eq sender operator)) true)
+		(let ((allowance (get-allowance sender operator)))
+			(asserts! (>= allowance amount) false)
+			(and
+				(< allowance max-uint128)
+				(map-set allowances {sender: sender, operator: operator} (- allowance amount))
+			)
+			true
```
friedger, line 39: Why the extra `true`?

sbtc-mini/contracts/sbtc-withdrawal-request-stx.clar
```diff
@@ -0,0 +1,68 @@
...
+	)
+)
+
+(define-private (is-allowed-operator-and-deduct (sender principal) (operator principal) (amount uint))
+	(begin
+		(asserts! (not (is-eq sender operator)) true)
+		(let ((allowance (get-allowance sender operator)))
+			(asserts! (>= allowance amount) false)
+			(and
+				(< allowance max-uint128)
```
friedger, line 36: Is this check necessary? The type checker would do it, wouldn't it?

sbtc-mini/contracts/sbtc-withdrawal-request-stx.clar
```diff
@@ -0,0 +1,68 @@
...
+		;; Check if the operator is allowed to request a withdrawal for sender for the specified amount.
+		(asserts! (is-allowed-operator-and-deduct sender contract-caller amount) err-operator-not-allowed)
+		;; Lock the tokens.
+		(unwrap! (contract-call? .sbtc-token protocol-lock amount sender) err-token-lock-failed)
+		;; Insert the request, returns the withdrawal request-id.
+		(contract-call? .sbtc-registry insert-withdrawal-request amount sender (get-expiry-burn-height) destination 0x)
+	)
+)
+
+(define-public (request-withdrawal-sponsored (amount uint) (sender principal) (destination { version: (buff 1), hashbytes: (buff 32) }) (fee uint))
```
friedger, line 55: This function could transfer the fee and then call `request-withdrawal`.

The destination check should be in both functions.

sbtc-mini/contracts/sbtc-withdrawal-request-stx.clar
MarvinJanssen: The function should return `true` if the operator is allowed.

Conditions:
- The sender and operator are equal, short-circuit `true`.
- The allowance of the operator must be equal to or larger than the amount, short-circuit `false` otherwise.
- Allowance equal to `max-uint128` means infinite allowance. If it is less than that then we need to update the map. The `and` expression is just an if without an else branch.
- The `true` at the end is necessary because the `and` returns `false` if the allowance is equal to `max-uint128`.

sbtc-mini/contracts/sbtc-withdrawal-request-stx.clar
MarvinJanssen: Which check? See the previous comment.

sbtc-mini/contracts/sbtc-withdrawal-request-stx.clar
MarvinJanssen: Will change.

sbtc-mini/contracts/sbtc-withdrawal-request-reclaim.clar
MarvinJanssen: Good question. It is hard to decide how opinionated we want the registry to be. I feel like it is too opinionated already.

What are your thoughts?

sbtc-mini/contracts/sbtc-withdrawal-request-stx.clar
friedger: I see. I was missing the special meaning of max-unit128.
Best to add a comment.

sbtc-mini/contracts/sbtc-withdrawal-request-reclaim.clar
friedger: I don't have an opinion. Light weight registery sounds good, holding only state. Out of scope for this PR

sbtc-mini/contracts/sbtc-withdrawal-request-stx.clar
MarvinJanssen: I added it to simplify the implementation but perhaps the `max-uint128` thing goes against the ethos of Clarity. What do you think?

sbtc-mini/contracts/sbtc-withdrawal-request-reclaim.clar
MarvinJanssen: Need to open an issue to discuss. I will write it tomorrow. Will keep this conversation open until the issue is made and linked.

sbtc-mini/contracts/sbtc-withdrawal-request-stx.clar
friedger: I would have choosen u0 or none as unlimited allowance.

MarvinJanssen: `u0` is dangerous I feel. `none` also feels like nothing but I think I would prefer it.

MarvinJanssen: I will keep `max-uint128` for now.

sbtc-mini/contracts/sbtc-withdrawal-request-reclaim.clar
MarvinJanssen: Here it is: #682

sbtc-mini/contracts/sbtc-token.clar
```diff
@@ -130,6 +130,10 @@
 	(ok (+ (ft-get-balance sbtc-token who) (ft-get-balance sbtc-token-locked who)))
 )
 
+(define-read-only (get-balance-available (who principal))
```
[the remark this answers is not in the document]
MarvinJanssen: This will no longer be necessary given the pivot of mini.

MarvinJanssen:
Good to go @friedger

sbtc-mini/contracts/clarity-bitcoin.clar
```diff
@@ -555,7 +555,8 @@
     ;; verify that the coinbase tx is correct
     (try! (was-tx-mined-compact burn-height ctx header { tx-index: u0, hashes: cproof, tree-depth: tree-depth }))
     (let (
-      (witness-out (get-commitment-scriptPubKey (get outs (try! (parse-tx ctx)))))
+      (witness-out (get-commitment-scriptPubKey (if (and (is-eq (element-at? ctx u4) (some 0x00)) (is-eq (element-at? ctx u5) (some 0x01))) (get outs (try! (parse-wtx ctx))) (get outs (try! (parse-tx ctx))
```
friedger, line 558: We can't accept wtx here for coinbase tx because we use `was-tx-mined-compact` before.

We could add a function that accepts coinbase txs with witnesses. However, I think we should limit the onchain work to what is necessary.

friedger: The test fails when we revert this change because the block header contains the wrong merkle root.

setzeus (approved):
Pulled & tested, can confirm all 100 unit tests working - LGTM.

PR #602, opened 2023-07-22
feat: Make alpha coordinator address recovery Hiro compatible

### LINKED ISSUE #600: Make alpha coordinator withdrawal signature recovery compatible with Hiro wallet message signing
Problem: When creating withdrawal payloads with the Hiro wallet, we need the wallet to sign a payload. However, the Hiro wallet message signing prefixes the messages signed with a hard-coded string plus the length of the message encoded as a varuint (https://github.com/bitcoinjs/varuint-bitcoin). This makes it impossible for the sBTC bridge (or any other app) to create withdrawal transactions for sBTC alpha.

Long term, we need to feature request arbitrary message signing without prefixing for the Hiro Wallet. However, a dirty quick fix to get alpha working is just to mimic the prefix in the coordinator. This should be quick to try out and verify, so to enable further debugging of alpha short term we'll create a branch with this modification.

### Definition of done
The alpha coordinator recovers signatures in a way compatible with how the Hiro wallet signs things.

netrome:
closes #600

### DIFF
--- a/.github/workflows/clarinet.yml
+++ b/.github/workflows/clarinet.yml
@@ -2,8 +2,10 @@ name: Clarinet
 on:
   push:
     branches: [ "main" ]
+    paths-ignore: ['**.md']
   pull_request:
     branches: [ "main" ]
+    paths-ignore: ['**.md']
 
 jobs:
   check:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -31,7 +31,7 @@ jobs:
     strategy:
       matrix:
         # can't run on `windows-latests`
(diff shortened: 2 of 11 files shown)


netrome:
I'll resurrect this PR and accommodate the new learnings from my session with @radicleart:

Some noteworthy quirks to accomodate in how the wallet does things:
 - It puts the recovery ID last in the signature.
 - It adds a hard-coded prefix to the message before hashing and signing.
 - It also adds the length of the message as a bitcoin varuint encoded prefix to the message.
 - It takes a message string as input and encodes it as utf8. We'd like to be able to pass an arbitrary byte array, or a hex encoded string.

netrome:
Note: We need to keep this branch around for reference for a while at least.

PR #601, opened 2023-07-22
Temporarily disable --release option in rust matrix workflow to avoid…

AshtonStephens:
**Disable `--release` flag in rust github actions workflow.**

For details and justification see https://github.com/Trust-Machines/stacks-sbtc/issues/599.

Issue #466, opened 2023-06-20
CI hiccup: test (macos-latest) is broken


netrome:
The `macos-latest` tests are failing in CI, but this is not reproducible locally.

Example errors:
https://github.com/Trust-Machines/core-eng/actions/runs/5294893814/jobs/9593253073?pr=459
https://github.com/Trust-Machines/core-eng/actions/runs/5313903835/jobs/9634089890#step:11:1348

Error message:
```
Error: Invalid port specified in -bind: '0.0.0.0:0'
test frost_btc has been running for over 60 seconds
test frost_btc ... FAILED
```

### Mitigation plan:
1. Disable the macos-latest target: https://github.com/Trust-Machines/core-eng/pull/467
2. Fix the problem
3. Re-enable the target

### Definition of done
The CI bug is fixed

AshtonStephens:
I'm taking this over - it's a good ticket to get familiar with the testing practices of this repo. Hoping to have a good solution by Friday.

AshtonStephens:
This port is specified in [test-utils/src/lib.rs#L114](https://github.com/Trust-Machines/stacks-sbtc/blob/e2b21b4cbca1380fb573d5a02c5da99c8f74bc13/test-utils/src/lib.rs#L114) which is almost certainly not right, but is for some reason working with all the other environments. The odder thing is that when I change the port to something more correct based on the [min and max port](https://github.com/Trust-Machines/stacks-sbtc/blob/e2b21b4cbca1380fb573d5a02c5da99c8f74bc13/test-utils/src/lib.rs#L40-L41), like `20001` I get this error: 

```
 Error: Unable to bind to 0.0.0.0:20001 on this computer. Bitcoin Core is probably already running.
 Error: Failed to listen on any port. Use -listen=0 if you want this.
```

Looking into what might be the issue.

CharlieC3:
You may already know this but a quick look through suggests macos/darwin does not map port 0 to mean any port, so it's being interpreted literally. On other OS's, it seems this is a shortcut to let the OS pick a random port not being used at the moment, thus avoiding any conflicts.
For this to work on macos, I'm guessing you'll either need to figure out if macos has another way of telling it to use any port aside from specifying port 0. Alternatively, you can rework the test logic to find an empty port on the current machine and use that.

PR #603, opened 2023-07-22
Explicitly assign ports to bitcoind command in testing. (#466)

AshtonStephens:
# Explicitly Choose Available Port when Instantiating `BitcoinProcess` &  Enable `macos-latest` Github Workflow Matrix Item

This change includes an explicit port choosing procedure that prevents other BitcoinProcess instances from claiming the same port before spawning an instance that actually listens on the port. This change includes a class `ClaimedPortFactory` that tracks which ports have been claimed and ensures that no other BitcoinProcess class instances can claim the same port until the BitcoinProcess that claimed that port has been dropped.

This is a more robust method preventing collisions what we were doing before because it makes no assumptions about whether multiple `BitcoinProcess` instances are being created at the same time or whether they'll be spawned soon after the `BitcoinProcess` instance is created.

**Note:** I've removed the `--release` matrix option in this change to show that the pipeline is unclogged after this is approved, but will remove it from this commit once it's approved in favor of the pull request https://github.com/Trust-Machines/stacks-sbtc/pull/601.

Bonus changes:
- `/tmp/` file now created with uuid as opposed to our own random string implementation


test-utils/src/lib.rs
```diff
@@ -247,7 +285,7 @@ pub fn generate_wallet(
     let keypair = KeyPair::new(&secp, &mut rand::thread_rng());
 
     let secret_key = keypair.secret_key();
-    let private_key = PrivateKey::new(secret_key, Network::Regtest);
+    let private_key: PrivateKey = PrivateKey::new(secret_key, Network::Regtest);
```
AshtonStephens, line 288: Not sure how this happened but I'll remove once this has been approved.

.github/workflows/rust.yml
```diff
@@ -29,8 +29,8 @@ jobs:
     strategy:
       matrix:
         # can't run on `windows-latests`
-        os: [ubuntu-latest, buildjet-4vcpu-ubuntu-2204-arm]
-        type: ['', '--release']
+        os: [ubuntu-latest, buildjet-4vcpu-ubuntu-2204-arm, macos-latest]
+        type: [''] # Disabling --release here while getting approval but really this change should come from https://github.com/Trust-Machines/stacks-sbtc/pull/601
```
AshtonStephens, line 33: I've removed the `--release` flag because it's important that we see the testing for this ticket pass, but I'll remove this in the rest revision assuming that this one is approved. The actual change to remove the `--release` flag temporarily can be found here (https://github.com/Trust-Machines/stacks-sbtc/pull/601)

AshtonStephens:
I think it's not running the `buildjet-4vcp-ubuntu-2204-arm` because I price capped my github account. We should find a way to make our testing cheaper.

jferrant (approved):
Looks good! Nice and clean

stjepangolemac (commented):
As far as I know we run these tests in parallel, which means we can't rely on a mechanism like this to assign ports.

Also, I'm pretty sure `0` port will assing a free port randomly on MacOS too. I just ran this locally to confirm.

```
fn main() {
    let foo = std::net::TcpListener::bind("0.0.0.0:0").unwrap();
    dbg!(foo.local_addr().unwrap().port());
}
```

AshtonStephens:
> As far as I know we run these tests in parallel, which means we can't rely on a mechanism like this to assign ports.
> 
> Also, I'm pretty sure `0` port will assing a free port randomly on MacOS too. I just ran this locally to confirm.
> 
> ```
> fn main() {
>     let foo = std::net::TcpListener::bind("0.0.0.0:0").unwrap();
>     dbg!(foo.local_addr().unwrap().port());
> }
> ```

Yeah, to clarify this issue occurs within the github MacOS environment, so a local test won't be able to recreate this. To the point of the tests running in parallel, note the mutex on the global class. It can only be modified by one thread at a time, so this logic is thread safe.

stjepangolemac:
Sorry, I wasn't clear enough. When you run `cargo test` or `cargo nextest` cargo will run multiple tests in separate processes, so mutexes won't help us in this case.

Looking at the issue right now the error seems to be:

```
Error: Invalid port specified in -bind: '0.0.0.0:0'
```

Maybe it's a matter of not specifying a port in this argument? What happens if we set `-bind` to `0.0.0.0` and `-rpcport` to `0`?

stjepangolemac:
I had a chat with @AshtonStephens, some conclusions:

- I approve the changes in the `.github/` and `./test-util` directories
- The PR as it is has a lot of unrelated changes in it, that needs to be addressed
- I'd also like to see "pick a random port from range" approach tested as it might allow us to get rid of shared state and might be much simpler

sergey-shandar (approved):

test-utils/src/lib.rs
```diff
@@ -36,27 +34,68 @@ use nix::unistd::Pid;
...
+
 const BITCOIND_URL: &str = "http://abcd:abcd@localhost";
 const MIN_PORT: u16 = 20000;
 const MAX_PORT: u16 = 25000;
 
+// Lazily instantiate a static port factory that will only supply ports that haven't already been
+// claimed.
+//
+// Note: This structure will be insufficient if multiple rust files assign ports.
+static CLAIMED_PORT_FACTORY: Mutex<Lazy<ClaimedPortFactory>> =
```
sergey-shandar, line 49: I would try to avoid singleton by using the same object PortFactory for all our applications.

jo-tm (commented):
LGTM

PR #605, opened 2023-07-24
feat: generate error table in README.md from Clarity source files

MarvinJanssen:
While putting together the sheet with all error codes and descriptions, I realised that at some point it will all have to be converted to a Markdown table for the README. I like a single source of truth so I thought why not generate the table directly from the source file.

This PR works but is also a proposal. What it does it read all contract source via a `clarinet run` script, identify all error constants and write them to a table. (They must start with `err-`.) Comments behind the constant definitions are extracted as the descriptions. You can see the result of the script in the committed README.md.

Todos:
- [x] Detect error code reuse and fail.
- [x] Prepare the script for use in CI.
- [x] Add a GitHub Actions check that fails if the generated README.md is not equal to the committed README.md.

@setzeus @friedger @FriendsFerdinand @jo-tm what do you think?


friedger:
The most prominent use case I can think of is that users receive an error code and want to understand the meaning of it. Therefore, the table could be grouped by contract and have the numbers in the second column.

For the use case of error code clash detection this is nice. 

Maybe we can use the mermaid graph to understand if error codes flow to different contracts and find out which error code are eventually visible to the end user of sBTC.

FriendsFerdinand:
Yes, I completely agree.

MarvinJanssen:
>The most prominent use case I can think of is that users receive an error code and want to understand the meaning of it. Therefore, the table could be grouped by contract and have the numbers in the second column.

By "users", do you mean downstream developers or users using sBTC?

Once we have numbered all errors properly, they should automatically be grouped together. Or do you mean creating a table for each?

>Maybe we can use the mermaid graph to understand if error codes flow to different contracts and find out which error code are eventually visible to the end user of sBTC.

That is a good idea. Perhaps also worth discussing what kind of errors we want to expose to the outside.

MarvinJanssen:
I merged #619 and ran the script. Looks decent to me. The script now also exits with a non-zero code if it detects a duplicate error value. We could technically already add that to the CI.

MarvinJanssen:
Tried it in CI, but unfortunately it always passes because exit codes raised in `clarinet run` scripts are not propagated. See https://github.com/hirosystems/clarinet/issues/1107.

MarvinJanssen:
Ready for review.

sbtc-mini/tests/sbtc-stacking-pool_signer_test.clar
```diff
@@ -16,7 +16,7 @@
 			((registration-result
 				(contract-call? .sbtc-stacking-pool signer-pre-register u1000 mock-pox-reward-wallet-1)))
 			(asserts! (is-err registration-result) registration-result)
-			(asserts! (is-eq (unwrap-err! registration-result registration-result) u1) registration-result))
+			(asserts! (is-eq (unwrap-err! registration-result registration-result) u6001) registration-result))
```
MarvinJanssen, line 19: Nit: these error values should not be inline, this change came in from a merge. Let's address in a future PR.

sbtc-mini/tests/sbtc-stacking-pool_signer_test.clar
```diff
@@ -25,10 +25,10 @@
 		(result (contract-call? .sbtc-peg-transfer relay-handoff-fulfillment u2101 0x 0x u1 u1 (list) 0x 0x 0x (list))))
 		(asserts! (is-err result)
 			(err "Should return error"))
-		(asserts! (is-eq (unwrap-err-panic result) u0)
+		(asserts! (is-eq (unwrap-err-panic result) u7000)
```
MarvinJanssen, line 28: As above.

sbtc-mini/tests/sbtc-stacking-pool_signer_test.clar
```diff
@@ -25,10 +25,10 @@
...
 			(err (concat "Should return err u0, not " (error-to-string (unwrap-err-panic result)))))
 		(ok true))))
 
 ;; errors from sbtc-peg-transfer
 (define-private (error-to-string (error uint))
-	(unwrap! (element-at? (list "err-current-pool-not-found") error) "unknown error"))
\ No newline at end of file
+	(unwrap! (element-at? (list "err-current-pool-not-found") (- error u7000)) "unknown error"))
```
MarvinJanssen, line 34: As above.

friedger (approved):
LGTM

Nice work

FriendsFerdinand (approved):

MarvinJanssen:
1. Great. I did not plan to touch other people's error codes at first but I decided to merge @friedger's PR on that so I could test the `EXTRACT_CHECK` flag and CI. 
2. I would say go ahead, would be awesome to have the mermaid grape on top of the README.

PR #619, opened 2023-07-28
Update error codes

friedger:
This PR
* fixes #618 
* replaces error codes with 4 digit codes using specific following format:

Component | Code | Description
-- | -- | --
Controller | 1XXX | Errors related to the controller and upgrades.
Registry | 2XXX | Errors coming directly from the registry.
Token | 3XXX | Errors coming directly from the token.
Deposit | 4XXX | Errors related to making and processing BTC deposits.
Redemption | 5XXX | Errors related to redeeming sBTC for BTC.
Pool | 6XXX | Errors related to the sBTC stacking pool.
Hand-off | 7XXX | Errors related to the peg hand-off process.
Bitcoin library | 8XXX | Errors coming directly from the Bitcoin library / helper.
Debug | 9XXX | Debugging related errors.

* does not update bitcoin library to make merging #516 easier

PR #627, opened 2023-07-29
Indefinite delegation

### LINKED ISSUE #626: [stacking-pool] Delegation should be indefinite
**Describe the bug**
Currently, pre-signers delegate until the current block height

**Expected behavior**
Pre-signers delegate to the pool indefinitely.

friedger:
This PR
* set delegation for pre-signers to indefinite
* fixes #626 
* exposes pox errors as error codes with 600xx, 6000xx and 6000xx (the later in the function call the more zeros after the 6)
* removes spaces at the end of lines
* uses pox-3.get-stx-account for mocking

builds on top of #619


MarvinJanssen (commented):
Interesting approach. Why more zeroes? Surely we can fit them all in the `6xxx` space?

friedger:
@MarvinJanssen How to distinguish between the different errors within one function call like in signer-pre-register. We have three calls to pox-3. The additional zeros help us to identify which pox call has thrown the error. The last digits in the error can be looked up in pox-3 and we know exactly what happened.

MarvinJanssen:
I understand, can you not unwrap the error code and increment? In any case, I am fine with either.

friedger:
@MarvinJanssen not sure what you mean. This PR unwraps and increments by 60000. Do you want to increment by a different number?

MarvinJanssen:
What I meant is we could increment by a number while keeping the error in the 6xxx range.

friedger:
Something like space 60xx  until 65xx is for non-pox error, and then 66xx for first pox call, 66xx for second pox call and 67xx. We would ignore the pox errors 254 and 255.

friedger:
I have changed the base to #605

I'll create a new issue for the error handling and the nit in #605

MarvinJanssen:
Thanks!

friedger:
Fixed by https://github.com/Trust-Machines/stacks-sbtc/commit/2cb2e09da825e9d177e1e345ebc1f122dc3dbbb8

PR #629, opened 2023-07-31
Make signer-minimal a constant

### LINKED ISSUE #563: [sbtc-mini] sbtc-stacking-pool signer minimal design
**signer-minimal is 0**

It's not possible to pre register (`signer-pre-register`) atm because variable signer-minimal is 0 and has no setter.

It's possible to pass amount-ustx=0 but the method calls `contract-call? .pox-3 delegate-stx` and so passes 0 to pox-3 - not sure this make sense?

friedger:
This PR
* makes signer minimal in stacking-pool a constant
* sets the minimum to 10k STX
* fixes #563


FriendsFerdinand (approved):

MarvinJanssen (approved):
LGTM

PR #650, opened 2023-08-02
Add rustdoc enforcement to "mini" packages (#649)

AshtonStephens:
This change adds rustdoc enforcement to the following packages:

- `stacks-coordinator-mini`
- `stacks-signer-mini`

Additionally, it adds both of them as packages within the root `Cargo.toml` file.


friedger (approved):
Looks good to me

AshtonStephens:
My mistake - I lost track of this PR and should have been on it.

I'm closing it for now. There's not much work here and we can easily do this again when we decide to add docs.

PR #659, opened 2023-08-04
Activation state of stacking pool

### LINKED ISSUE #622: [stacking-pool] delegate-stack-increase called after commit
**Describe the bug**
in `signer-pre-register`, function `delegate-stack-increase` is called after stack-aggregation-commit. Increase should be called after extend if needed.

### LINKED ISSUE #625: [stacking-pool] add active state
**Is your feature request related to a problem? Please describe.**
After deployment and upgrade of the sbtc contracts, the pool is not active. Currently, there is no way to register as a signer in this state. 

**Describe the solution you'd like**
As described in https://github.com/stacks-network/sbtc-docs/issues/13#issuecomment-1656586097 add rules 
- [x] that the pool is active if and only if 1m stx or more are locked 
- [x] that users can read the active state. 
- [x] that the minimum of locked stx for signers is 10k stx.

friedger:
This PR
* adds a getter for active state
* prevents extending the stacking pool if the locked amount is below 1m STX
* fixes #625 
* fixes #622


sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -23,11 +23,15 @@
...
 ;; Dust limit placeholder for checking that pox-rewards were disbursed (in sats)
 (define-constant dust-limit u100)
+
 ;; Signer minimal is 10k STX
 (define-constant signer-minimal u10000000000)
 
 (define-constant pox-info (unwrap-panic (contract-call? .pox-3 get-pox-info)))
 
+;; Minimum amount of 1m locked STX for the pool to be active
+(define-constant minimal-pool-amount-for-activation u1000000000000)
```
MarvinJanssen, line 33: I wonder if we ever want to change this value. We could make it updatable by protocol call, then a future version could change it if need be.

MarvinJanssen (changes requested):
Looks good. Just need some unit tests for the new state check.

sbtc-mini/contracts/sbtc-stacking-pool.clar
friedger: Done

FriendsFerdinand:
Why are we going back to the `pox-3` for cycle-height related queries?

friedger:
@FriendsFerdinand What? Where?

friedger:
@MarvinJanssen Unit tests added

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -353,27 +367,19 @@
 			error (asserts! false (err (+ (to-uint error) const-second-pox-error))))
 
 		;; Stack aggregate-commit
-		;; As pointed out by Friedger, this fails when the user is already stacking. Match err-branch takes care of this with stack-delegate-increase instead.
+		;; This fails when the user is already stacking.
+		;; It does not matter because stx are already locked and ready to be extended in the next cycle
 		(match (as-contract (contract-call? .pox-3 stack-aggregation-commit-indexed pox-addr next-cycle))
```
setzeus, line 372: Is the (match ...) needed since both branches return true?

setzeus (approved):
Have a comment on a possibly-superfluous (match ...) but outside of that pulled, ran clarinet console & can confirm that all unit tests are passing. 

LGTM.

sbtc-mini/contracts/sbtc-stacking-pool.clar
friedger: Is there a different way to ignore the result? We are happy with ok or err

friedger: @setzeus ☝️

FriendsFerdinand (approved):

PR #663, opened 2023-08-07
[sBTC-mini] Unchecked raw-coinbase-tx

CAGS295:
Small issue I noted. It looks like the parsing checks are missing `raw-coinbase-tx`.


friedger:
It is not good that the change did not effect the tests. @CAGS295 

Either we should remove the contract call or add a check for the result and compare `parsed-coinbase-tx` with some data

friedger (changes requested):
Add an assert for `parsed-coinbase-tx`

CAGS295:
I agree, `.clarity-bitcoin was-segwit-tx-mined-compact` parses it internally [here](https://github.com/Trust-Machines/stacks-sbtc/blob/main/sbtc-mini/contracts/clarity-bitcoin.clar#L558) anyway. I lean towards removing the call and implementing a separate test for parse-tx instead.

FriendsFerdinand (approved):

PR #670, opened 2023-08-09
Add flow test with two signers, fix stacking pool

### LINKED ISSUE #669: [stacking-pool] Extend is incorrect
**Describe the bug**
Extending the current signer 
* does not use `as-contract`
* does not work because delegate-stx uses a fixed pox-address

friedger:
This PR
* makes stackers to delegate without a specific pox-address and thereby allows to switch stacking to different sbtc wallets.
* adds `as-contract` wrapping to delegate-stack-extend in vote function.
* fixes #669 
* adds a flow test with two stackers
* increases code coverage of stacking pool to 61%


sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -357,7 +357,7 @@
 		(asserts! (is-eq (get-current-window) registration)  err-not-in-registration-window)
 
 		;; Delegate-stx to their PoX address
```
setzeus, line 359: Clarifying for own sake here, why changing to none - since this is for pre-registration shouldn't they submit their own reward-address?

I recall Jude mentioning the following in the Master sBTC Document (could totally be outdated, just wanted to surface):

"They delegate to the sbtc contract (but supply a PoX reward address of their choosing)"

sbtc-mini/contracts/sbtc-stacking-pool.clar
```diff
@@ -543,40 +543,45 @@
...
 					(asserts!
 						(and
 							;; Assert that new-candidate-votes is greater than or equal to 70% of next-pool-total-stacked
 							(>= (/ (* new-candidate-votes u1000) next-pool-total-stacked) (var-get threshold-consensus))
 
 							;; Assert that extend-and-commit-bool is true (both delegate-stack-extend & aggregate-commit-indexed succeeded)
-							(match (fold mass-delegate-stack-extend next-pool-stackers (ok {stacker: tx-sender, unlock-burn-height: u0, pox-addr: pox-addr}))
-								passed-result
+							(match (as-contract (fold mass-delegate-stack-extend next-pool-stackers (ok {stacker: tx-sender, unlock-burn-height: u0, pox-addr: pox-addr})))
+								success-extend
```
setzeus, line 554: Great update, much better readability with success-extend & success-commit > passed-result & ok-result. 

Nothing to resolve.

sbtc-mini/contracts/sbtc-stacking-pool.clar
friedger: Signer delegates to the pool address and does not require that the pool operator uses one specific pox address (it will change every cycle). Therefore it should be none for `delegate-stx` and the signer's pox address for `delegate-stack-stx`. Later the pox address can be changed because the delegate call was with none as pox address.

setzeus (approved):
Pulled, reviewed & can confirm all 83 unit tests intact. LGTM!

FriendsFerdinand (approved):

jo-tm:
@friedger there is a conflict you can solve locally with VSCode or other:

image

jo-tm:
@friedger you can solve the conflicts in Github now too in this case, it seems. 
image

PR #671, opened 2023-08-10
Add unit tests with test contract as protocol contract

friedger:
This PR
* adds a unit test contract as protocol contract for unit tests
* adds tests to update threshold percentage and minimum stacking amount for pool


setzeus (approved):
Reviewed new unit tests which covers the following behaviors as the protocol call interacting with the stacking pool:

1. Protocol can update consensus/threshold percent (min of 49.9% & max of 95.1%)
2. Protocol can update the minimal amount of STX required (50k stx required)

All unit tests currently passing. LGTM.

sbtc-mini/tests/sbtc-stacking-pool_as_protocol_test.clar
```diff
@@ -0,0 +1,61 @@
+
+;; .sbtc-stacking-pool_as_protocol_test has been added as protocol contract
+;; in bootstrap file for unit testing.
+
+(define-constant err-error-expected (err u9901))
```
jo-tm, line 5: maybe we should separate the error codes on a separate contracts or library? so we can use them in the tests too? is this a good idea?

jo-tm (approved):
Approved but left a comment about error codes.

FriendsFerdinand (approved):

sbtc-mini/tests/sbtc-stacking-pool_as_protocol_test.clar
FriendsFerdinand: I don't think it's necessary just because these error codes are created at a file-level

sbtc-mini/tests/sbtc-stacking-pool_as_protocol_test.clar
```diff
@@ -0,0 +1,61 @@
+
+;; .sbtc-stacking-pool_as_protocol_test has been added as protocol contract
```
MarvinJanssen, line 2: I think I missed this but @friedger you added `.sbtc-stacking-pool_as_protocol_test` to the bootstrap file right?

It should not go there, you can add it as a protocol contract via a test prepare function.

Example:
https://github.com/Trust-Machines/stacks-sbtc/blob/d0cde0c23dd4fd381e0480f47a979d195ebb66fe/sbtc-mini/tests/sbtc-peg-out-verifier_test.clar#L28-L30

sbtc-mini/tests/sbtc-stacking-pool_as_protocol_test.clar
friedger: It is  not of a hassle to copy the error codes over. For now, I don't see a better solution

jo-tm: gotcha

sbtc-mini/tests/sbtc-stacking-pool_as_protocol_test.clar
friedger: Good to now! I remove the unit test from the bootstrap file @MarvinJanssen

PR #678, opened 2023-08-16
fix: change burn-block-height to 3702

friedger:
This PR
* re-enables a flow test without change
* fixes a unit test by increasing the expected burn-block-height (not sure where the extra block comes from)


friedger:
Result of running tests.sh @MarvinJanssen 
image

MarvinJanssen:
LGTM

PR #697, opened 2023-09-08
Asyncify the repository

tbraun96:
Using blocking code is an old pattern that is susceptible to the c10k problem, especially since we are in a networking context. We should default to async code to speed up performance.

* Use sqlx over rusqlite
* Use reqwest over ureq
* Use tokio::sync::mpsc::UnboundedSender/Receiver over std::sync::mpsc::Sender/Receiver

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,6 @@ hashbrown = { version = "0.14", features = ["serde"] }
 bincode = "1.3.3"
 itertools = "^0.11.0"
 sha2 = "0.10"
-ureq = { version = "2.6", features = ["json"] }
 url = "2.3.1"
 rand = "0.8.5"
 backoff = "0.4"
@@ -47,3 +46,6 @@ aes-gcm = "0.10"
 sha256 = "=1.4.0"
 secp256k1 = { version = "0.27.0", features = ["global-context", "recovery", "serde", "bitcoin-hashes"] }
 regex = "~1.8.4"
+tokio = "1.32.0"
+ureq = { version = "2.7.1", features = ["json"]}
+async-trait = "0.1.73"
--- a/frost-signer/src/main.rs
+++ b/frost-signer/src/main.rs
@@ -5,7 +5,8 @@ use frost_signer::config::{Cli, Config};
 use frost_signer::logging;
 use frost_signer::signer::Signer;
 
-fn main() {
+#[tokio::main]
+async fn main() {
     logging::initiate_tracing_subscriber();
 
     let cli = Cli::parse();
@@ -20,7 +21,7 @@ fn main() {
             ); // sign-on message
 
             //Start listening for p2p messages
-            if let Err(e) = signer.start_p2p_sync() {
+            if let Err(e) = signer.start_p2p_async().await {
                 warn!("An error occurred in the P2P Network: {}", e);
             }
         }
--- a/frost-coordinator/Cargo.toml
+++ b/frost-coordinator/Cargo.toml
@@ -19,6 +19,7 @@ tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 frost-signer = { path = "../frost-signer" }
 serde = { version = "1.0", features = ["serde_derive"] }
+tokio = { workspace = true, features = ["full"] }
 
 [dev-dependencies]
 rand_core = { workspace = true }
--- a/frost-coordinator/src/coordinator.rs
+++ b/frost-coordinator/src/coordinator.rs
@@ -50,7 +50,7 @@ pub enum Command {
 }
 
 pub struct Coordinator<Network: NetListen> {
-    id: u32, // Used for relay coordination
+    id: Network::Arg, // Used for relay coordination
     current_dkg_id: u64,
     current_dkg_public_id: u64,
     current_sign_id: u64,
@@ -68,7 +68,7 @@ pub struct Coordinator<Network: NetListen> {
 }
 
 impl<Network: NetListen> Coordinator<Network> {
-    pub fn new(id: u32, config: &Config, network: Network) -> Result<Self, Error> {
+    pub fn new(id: Network::Arg, config: &Config, network: Network) -> Result<Self, Error> {
         Ok(Self {
             id,
             current_dkg_id: 0,
@@ -113,20 +113,20 @@ impl<Network: NetListen> Coordinator<Network>
 where
     Error: From<Network::Error>,
 {
-    pub fn run(&mut self, command: &Command) -> Result<(), Error> {
+    pub async fn run(&mut self, command: &Command) -> Result<(), Error> {
         match command {
             Command::Dkg => {
-                self.run_distributed_key_generation()?;
+                self.run_distributed_key_generation().await?;
                 Ok(())
             }
             Command::Sign { msg } => {
-                self.sign_message(msg)?;
+                self.sign_message(msg).await?;
                 Ok(())
             }
             Command::DkgSign { msg } => {
                 info!("sign msg: {:?}", msg);
-                self.run_distributed_key_generation()?;
-                self.sign_message(msg)?;
+                self.run_distributed_key_generation().await?;
+                self.sign_message(msg).await?;
                 Ok(())
             }
             Command::GetAggregatePublicKey => {
@@ -137,17 +137,17 @@ where
         }
     }
 
-    pub fn run_distributed_key_generation(&mut self) -> Result<Point, Error> {
+    pub async fn run_distributed_key_generation(&mut self) -> Result<Point, Error> {
         self.current_dkg_id = self.current_dkg_id.wrapping_add(1);
         info!("Starting DKG round #{}", self.current_dkg_id);
-        self.start_public_shares()?;
-        let public_key = self.wait_for_public_shares()?;
-        self.start_private_shares()?;
-        self.wait_for_dkg_end()?;
+        self.start_public_shares().await?;
+        let public_key = self.wait_for_public_shares().await?;
+        self.start_private_shares().await?;
+        self.wait_for_dkg_end().await?;
         Ok(public_key)
     }
 
-    fn start_public_shares(&mut self) -> Result<(), Error> {
+    async fn start_public_shares(&mut self) -> Result<(), Error> {
         self.dkg_public_shares.clear();
         info!(
             "DKG Round #{}: Starting Public Share Distribution Round #{}",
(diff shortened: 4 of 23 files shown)


tbraun96:
I have not ran the tests yet, however, all the tests compile and there are no lint errors.

xoloki:
I approved the workflows for CI, so you should get CI working now when pushing to your fork's PR branch.

I tested the changes locally, and unfortunately the frost-coordinator integration tests are hanging.  Run `cargo test` in the `frost-coordinator` subdir to try this yourself.  Once you get `integration_test_frost_coordinator_should_be_able_to_successfully_run_dkg_sign` running the rest should be easy.  

The other test that matters is the `frost_btc` test in the `frost-test` subdir, but it doesn't rely on the networking so it's still passing locally for me.  If you want to try it you will need bitcoind v22.0 (or greater) in your PATH.

Thanks so much for this PR!  It's a significant enhancement to the creaky networking the repo currently uses.  This repo was used for the initial alpha release of sBTC, and now it's mostly used as a PoC for external collaborators.  It would be excellent to get the networking up to par.

tbraun96:
Interestingly, when I run `cargo test` inside the `frost-coordinator` directory, the tests all pass. However, where I am getting some hanging behavior is inside the `stacks-coordinator` directory. Currently, these tests appear to be hanging:

```
test stacks_node::client::tests::burn_block_height_failure_test has been running for over 60 seconds
test stacks_node::client::tests::burn_block_height_success_test has been running for over 60 seconds
test stacks_node::client::tests::call_read_failure_test has been running for over 60 seconds
test stacks_node::client::tests::call_read_success_test has been running for over 60 seconds
test stacks_node::client::tests::keys_threshold_invalid_test has been running for over 60 seconds
test stacks_node::client::tests::keys_threshold_test has been running for over 60 seconds
test stacks_node::client::tests::next_nonce_failure_test has been running for over 60 seconds
test stacks_node::client::tests::next_nonce_success_test has been running for over 60 seconds
```

I will continue to investigate this.

tbraun96:
The tests now all pass for me locally. The exceptions are the `yarpc` tests, which should work in the pipeline.

xoloki:
> The tests now all pass for me locally. The exceptions are the `yarpc` tests, which should work in the pipeline.

Ah, the `frost-coordinator` tests do pass, they just take a very long time:
`
x@xocotl:~/src/stacks-sbtc-tbraun96/frost-coordinator$ time cargo test coordinator::test::integration_test_frost_coordinator_should_be_able_to_successfully_run_dkg_sign
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 10.32s
`
vs
`
x@xocotl:~/src/stacks-sbtc/frost-coordinator$ time cargo test integration_test_frost_coordinator_should_be_able_to_successfully_run_dkg_sign 
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.22s
`

So with your async code it's taking 50x longer for dkg/sign.  Can you investigate why that is?  

One obvious difference is that you're running `tokio::task::spawn` instead of `thread::spawn` for the signers.  Most of the time the signers use is time doing crypto, so they need to run in their own threads.  But that would only account for a 6x slowdown, not a 50x slowdown.  Can you configure a thread pool for `tokio::task`s?

tbraun96:
Here is my benchmark for `cargo test` against `frost-coordinator` (on a Mac M1 basic CPU):

```
test coordinator::test::integration_test_frost_coordinator_should_be_able_to_successfully_run_dkg_sign ... ok
test coordinator::test::integration_test_frost_coordinator_should_provide_valid_signatures_after_restart ... ok
test coordinator::test::integration_test_frost_coordinator_should_provide_valid_signatures_after_dkg ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.43s
```

When I switch to a multi-threaded runtime (see how I did this with the latest commit diff):

```
test coordinator::test::integration_test_frost_coordinator_should_be_able_to_successfully_run_dkg_sign ... ok
test coordinator::test::integration_test_frost_coordinator_should_provide_valid_signatures_after_dkg ... ok
test coordinator::test::integration_test_frost_coordinator_should_provide_valid_signatures_after_restart ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.87s
```

> One reason why I've never gotten in the habit of using async with Rust is that compared to C/C++, the APIs seem so primitive and badly constructed. In C/C++, for example, I would use poll/select to dispatch IO from a large set of file descriptors, but then I would use message queues with a thread pool to actually handle the IO. So there's no blocking IO, but I can still configure the exact amount of concurrent processing that I want for any given set of IO handlers.

We can do this in Rust too. I spent a couple years learning just how to write proper async code in rust, and can say now that it's basically second nature at this point. It's just a different language domain, but you can do pretty much everything you can do in other languages. It can be tricky at first though, but the steepness of that learning curve starts to vanish.

Check to see the performance on your machine with the latest commit.

xoloki:
@tbraun96 I ran just the `successfully_run_dkg_sign` test with your latest code, three times, and got the following:

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 43.55s
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 82.61s
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 51.34s

I then tried running all three `frost-coordinator` tests at once, and got:

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 254.77s

I verified that your change `#[tokio::test(flavor = "multi_thread")]` was present, and all cores on my machine were in use.  This is a Ryzen 5950X on Linux, with 16 fast cores.

I'll try again on my M2 MacBook Air and see if there's a difference.

What performance do you get when you run the original threaded/blocking code?  I consistently get around `0.25s` for the `successfully_run_dkg_sign` test.

xoloki:
@tbraun96 I tried the original threaded/blocking code on my M2 MBA, and consistently get around `0.35s` for the `successfully_run_dkg_sign` test.  

For some reason `cargo` is complaining about building your fork:

error: no matching package named `blockstack-core` found
location searched: https://github.com/stacks-network/stacks-blockchain/?branch=next
required by package `sbtc-cli v0.1.0 (/Users/axoloki/src/stacks-sbtc-tbraun96/sbtc-cli)`

I tried with both `https:` and `git:` checkouts.  Didn't see this problem on Linux.  I'll try updating `cargo`.

xoloki:
Weird, same problem after `rustup update`.  `cargo --version` is now `1.72.0` on mac and `1.70.0` on linux.  I'll try disabling the offending `sbtc-cli` package on mac.

xoloki:
After disabling `sbtc-cli` and `stacks-signer` it now runs.  `successfully_run_dkg_sign` now takes:

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.46s

All `frost-coordinator` tests together now take:

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.71s

Do you have a linux box to test on?  It seems bizarre that linux would be 3 orders of magnitude slower with a faster CPU :O

xoloki:
Oh, it looks like `stacks-network/stacks-blockchain` just merged their `develop` branch into `next`, so the `blockstack-core` package is gone.  Weirdly, using a cargo git dependency with a `rev` doesn't seem to fix it.

xoloki:
Ok, changing the `blockstack-core` dep's branch to `master` works.  You should probably do this on your fork.

xoloki:
So, apparently the fact that `stacks-blockchain` defined a top-level package of `blockstack-core` instead of `stacks-blockchain` confuses `cargo` sufficiently that a git dep with rev doesn't work in this context.

xoloki:
It wasn't a force push, it was just a non-squash merge of many many commits.  Change your  Cargo.toml to point to:

blockstack-core = { git = "https://github.com/stacks-network/stacks-blockchain/", rev = "edb4b7284a6d6c4326cf1d7204edb5756330aa46" }

I'll fix that upstream now.

xoloki:
Ok upstream fixed.  Let me know if you have any idea why it's slow on Linux.

xoloki:
We're still running slow on Linux.  Release is usually faster than debug, but still two orders of magnitude slower than Mac, or the threaded/blocking code on Linux:

x@xocotl:~/src/stacks-sbtc-tbraun96/frost-coordinator$ cargo test coordinator::test::integration_test_frost_coordinator_should_be_able_to_successfully_run_dkg_sign
     Running unittests src/lib.rs (/home/x/src/stacks-sbtc-tbraun96/target/debug/deps/frost_coordinator-0225c0f3b17d0d97)
test coordinator::test::integration_test_frost_coordinator_should_be_able_to_successfully_run_dkg_sign ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 71.16s

x@xocotl:~/src/stacks-sbtc-tbraun96/frost-coordinator$ cargo test --release coordinator::test::integration_test_frost_coordinator_should_be_able_to_successfully_run_dkg_sign
     Running unittests src/lib.rs (/home/x/src/stacks-sbtc-tbraun96/target/release/deps/frost_coordinator-cc3b59b0804dd424)
test coordinator::test::integration_test_frost_coordinator_should_be_able_to_successfully_run_dkg_sign ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 55.62s

I'm going to run the profiler on linux and see what the flamegraph tells me.  I'm guessing that we're doing a large number of null callbacks that's eating up 99% of the time.

xoloki:
The problem is that `reqwest` uses SSL by default, and on Linux it uses OpenSSL in a particularly stupid way.  It loads certs on every connection, and since we're opening new connections on every poll to the `relay-server`, we spend all of our time loading SSL certs.

Fortunately, there's an easy fix to disable SSL for `reqwest` in the root Cargo.toml:

-reqwest = "0.11.14"
+reqwest = { version = "0.11.14", default-features = false }

With this change in debug we get:
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.24s

And in release:
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 2 filtered out; finished in 0.12s

Make this change, rebase on upstream to pickup the various CI fixes, and we should be good.

frost-signer/src/main.rs
```diff
@@ -5,7 +5,8 @@ use frost_signer::config::{Cli, Config};
 use frost_signer::logging;
 use frost_signer::signer::Signer;
 
-fn main() {
+#[tokio::main]
```
xoloki, line 8: All of the `tokio::main` annotations should be multithreaded:

`#[tokio::main(flavor = "multi_thread")]`

Cargo.toml
```diff
@@ -47,3 +46,5 @@ aes-gcm = "0.10"
 sha256 = "=1.4.0"
```
xoloki, line 46: The reqwest line should be:

`reqwest = { version = "0.11.14", default-features = false }`

xoloki (changes requested):
Couple of changes to make, then rebase on upstream and I'll retest.  Should be a quick approval after that.

Thanks!

tbraun96:
Added an additional commit to allow the `Signer` implementation to take a custom NetListen implementation.

PR #699, opened 2023-09-11
fix: use new stackslib dependency

stjepangolemac:


xoloki (approved):
LGTM

xoloki:
Thinking about this some more, it's probably safer in terms of preventing this from happening again if we stay with the `rev` style git dependency.  Maybe change the `branch` specification to whatever the current `rev` is?  What does everyone think?

It was really surprising to find 3 different bit-rotted pieces of the build at the same time, and doesn't really inspire confidence from the external collaborators who use this build for integration.

jo-tm (approved):
the library change was straightforward, the rest are trailing spaces removed.

